PR #757, opened 2016-06-10
fix(core): set default language at configService init

### LINKED ISSUE #620: The map for index-fgp-fr does not load
Removing `rv-langs` lets the map load but we lose loading with french translations...
As a starting point the page works in `v0.14.0` even with `rv-langs='["fr-CA"]'`

ChrisLatRNCan:
RunBlock (core.run) execution set the language by default to en-CA
while the configService generate fr-CA as the only key
if rv-langs = '[fr-CA]' so a conflict is generate at configService.getCurrent() call
Solution: remove RunBlock language setting and set the config or default language
in configService right after the rv-langs value extraction

Closes #620

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/757)

### DIFF
--- a/src/app/core/config.service.js
+++ b/src/app/core/config.service.js
@@ -96,6 +96,9 @@
                         langs = ['en-CA', 'fr-CA'];
                     }
 
+                    // set the language right away and not wait the initialization to be fullfilled
+                    $translate.use(langs[0]);
+
                     langs.forEach(lang => partials[lang] = []);
 
                     if (svcAttr) {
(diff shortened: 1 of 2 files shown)


spencerwahl:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/757)*

PR #905, opened 2016-07-15
feat(ui): Add query off layer flag

### LINKED ISSUE #668: Add `query-off` flag to legend entries
Follow up on #409:

There is no visible indication on legend items that the `query` option is turned off right now.
We discussed it and think a good solution would be to display a `query-off` flag on the legend entry when it's turned off. Something like this:
2016-05-24 14_32_58-adobe photoshop cs3
This way it's easy to tell which layer or sublayer will not appear in the identify results. This will be a separate task.

Notes:
- the flag should be visible __only__ when the query is turned off for this particular legend entry
- the flag needs a meaningful tooltip (suggestions are welcome) because phrases like `query is disabled` or `identify is disabled` can be meaningless to the user

milesap:
Closes #668

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/905)

### DIFF
--- a/src/app/ui/settings/settings.directive.js
+++ b/src/app/ui/settings/settings.directive.js
@@ -101,7 +101,13 @@
 
             if (tocEntry.type === 'group') {
                 // set include subgroup to true to cascade to all (group, subgroups and items)
-                tocEntry.walkItems(item => item.options.query.value = value, true);
+                tocEntry.walkItems(item => {
+                    item.options.query.value = value;
+                    item.flags.query.visible = !value;
+                }, true);
+            } else {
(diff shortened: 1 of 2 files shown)


alyec:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/905)*

spencerwahl:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/905)*

Issue #834, opened 2016-06-24
Full page app support


spencerwahl:
Put a flag in the config that lets the viewer know its in control of the page. Control meaning either:
- the viewer is the only content on the page (as in `index-one.html`)
- or it is in full-screen mode within the host page

Update `focus.service.js` so that it uses this config value, currently we have a `rv-fullpage-app` attribute in `index-one.html` as a temporary solution.

Should turn on things like the share button.

AleksueiR:
This flag should be true when the viewer is __the only content on the page__ and when the viewer is in __the full-screen mode within the host page__.

This will also be used by the focus manager to not show the "keyboard focus" dialog when then viewer is in the full screen.

See PR #1167

PR #1199, opened 2016-09-19
fix(ui): focus manager correctly responds to mouse activity

### LINKED ISSUE #1192: Focus manager erroneously triggered
Steps:
- click on the map with a mouse 
- press "Tab" or "Shift + Tab"

"Keyboard focus" dialog gets triggered.

See PR fgpv-vpgf/fgpv-vpgf#1167.

Expected behaviour:
After clicking on the map, the focus should stay/be placed inside the viewer; pressing "Tab" should not trigger the dialog.

milesap:
Clicking inside the viewer with a mouse completely disables focus
management until focus leaves the viewer.

Closes #1192

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/1199)

### DIFF
--- a/src/app/layout/focus.service.js
+++ b/src/app/layout/focus.service.js
@@ -31,8 +31,8 @@
         let lastFocusElement;
         let firstFocusable; // first focusable element in the viewer
         let lastFocusable; // last focusable element in the viewer
-        let keyLastPressed; // boolean if a key rather than mouse was pressed/clicked
         let currentStatus; // current status of the focus manager
+        let isFullPage;
 
         const statuses = {
             NONE: undefined,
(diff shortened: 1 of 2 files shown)


AleksueiR:
This fixes #1192; I have a small concern however - by disabling the focus manager we break the keyboard navigation inside the viewer (after the user clicked with a mouse) until the focus leaves the viewer and the user tabs back in. In essence, we are creating two modes of keyboard navigation: one when tabbing into the viewer; another, when clicking it. This will work well under the assumption that users "activating" the viewer with a mouse will not use keyboards to interact with it. A user who wants to use keyboard navigation inside the viewer after clicking it might find it not working properly.

@mweech Are we okay with such an assumption?

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1199#-:-KS2JqtrCwU4JNuWNu6k:b-acowf9)*

milesap:
I made a small change to allow limited simultaneous mouse and keyboard support. In most cases it works well but a more formal update would be required to eliminate some minor issues. If it works well enough as-is we can leave it like this for now until a future update fully supports simultaneous mouse/keyboard navigation.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1199#-:-KS2RZ1oyIfEDUzJh1qE:b-8k78yq)*

milesap:
Added attribute to `index-one.html` for now until config has this property (issue #834). This also fixes the odd behaviour on `index-many.html` - the cause was indeed that the first map didn't see any focusable elements from the second or third map since they hadn't loaded yet. 

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1199#-:-KS6TEvVBqF4lZUMuoKB:b-hmvwky)*

AleksueiR:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1199)*

Issue #1482, opened 2016-12-07
Datatable "jumping" when scrolling and ordering columns


AleksueiR:
Set the column sizes to maximum sizes allowed by the service for corresponding fields (need a reasonable ceiling - 200 characters for example - in case the numbers are too big).

PR #1546, opened 2016-12-23
fix(filters): Datatables column jumping when sorting

jolevesq:
## Description

Datatable jump when sorting field. To solve this, we must set column width. For long text, we render it with ellipsis. If you hover it you will have the value. For wcah keyboard user can tab to the ellipsis to get the value. TEsted with http://geoappext.nrcan.gc.ca/arcgis/rest/services/GSCC/Geochronology/MapServer/ who has a lot of long fields. Issue #1482

## Testing
Chrome, FF and Safari

## Documentation
Inline>

## Checklist

- [x] code passes unit tests

### DIFF
--- a/src/app/ui/filters/filters-default.directive.js
+++ b/src/app/ui/filters/filters-default.directive.js
@@ -166,6 +166,31 @@
                         .filter(idx => idx > -1);
                 };
 
+                // set width from field length if it is a string field type. If it is the oid field,
+                // set width to 100px because we have the oid, the details and zoom to button. If it is
+                // another type of field, set width to be the title.
+                displayData.columns.forEach(column => {
+                    const field = displayData.fields.find(field => field.name === column.data);
+
+                    if (typeof field !== 'undefined') {
+                        if (field.type === 'esriFieldTypeString') {
+                            const width = getColumnWidth(column.title, field.length);
+                            column.width = `${width}px`;
+                            column.render = renderEllipsis(width);
+                        } else if (field.type === 'esriFieldTypeOID') {
+                            // set column to be 100px width because of details and zoom to buttons
+                            column.width = '100px';
+                        } else {
+                            const width = getColumnWidth(column.title);
+                            column.width = `${width}px`;
+                            column.render = renderEllipsis(width);
+                        }
+                    } else {
+                        // set symbol column width
+                        column.width = '30px';
+                    }
+                });
+
                 // ~~I hate DataTables~~ Datatables are cool!
                 self.table = tableNode
                     .on('init.dt', callbacks.onTableInit)
@@ -200,6 +225,91 @@
                         ],
                         oLanguage: oLang
                     });
+
+                /**
+                 * Get column width from column title and field length.
+                 * @function getColumnWidth
+                 * @private
+                 * @param {Object} title    column title
+                 * @param {Object} length   optional column length (characters)
+                 * @param {Object}  maxLength   optional maximum column length (pixels)
+                 * @return {Number} width    width of the column
+                 */
+                function getColumnWidth(title, length = 0, maxLength = 200) {
+                    // get title length (minimum 50px)
+                    let metricsTitle = getTextWidth(title);
+                    metricsTitle = metricsTitle < 50 ? 50 : metricsTitle;
+
+                    // get column length (only type string have length)
+                    if (length) {
+                        // generate a string with that much characters and get width
+                        let metricsContent = getTextWidth(Array(length).join('x'));
+
+                        // set the column length from field length (maximum will be maxLength)
+                        metricsContent = metricsContent <= maxLength ? metricsContent : maxLength;
+
+                        // check if it is lower then title length. If so, use title length
+                        metricsTitle = metricsContent < metricsTitle ? metricsTitle : metricsContent;
+                    }
+
+                    return metricsTitle;
+                }
+
+                /**
+                 * Render long text width ellipsis (https://datatables.net/blog/2016-02-26)
+                 * @function RenderEllipsis
+                 * @private
+                 * @param {Object} width    column width
+                 * @return {String} text    text for td element or string who contain html element
+                 */
+                function renderEllipsis(width) {
+                    const esc = (text) => {
+                        return text
+                            .replace(/&/g, '&amp;')
+                            .replace(/</g, '&lt;')
+                            .replace(/>/g, '&gt;')
+                            .replace(/"/g, '&quot;');
+                    };
+
+                    return (text, type) => {
+                        // order, search and type get the original data
+                        if (type !== 'display') {
+                            return text;
+                        }
+
+                        if (typeof text !== 'number' && typeof text !== 'string') {
+                            return text;
+                        }
+
+                        text = text.toString(); // cast numbers
+
+                        // if text width smaller then column width, return text
+                        if (getTextWidth(text) < width) {
+                            return text;
+                        }
+
+                        // for wcag we add a text input read only. This element is focusable so we can have tooltips.
+                        return `<input type="text" readonly title="${esc(text)}" value="${esc(text)}"
+                                    class="rv-render-ellipsis"></input>
+                                <span class="rv-render-tooltip">${esc(text)}</span>`;
+                    };
+                }
+
+                /**
+                 * Get text width (http://stackoverflow.com/questions/118241/calculate-text-width-with-javascript)
+                 * @function getTextWidth
+                 * @private
+                 * @param {String} input    text ot calculate width from
+                 * @return {Number} width    text width
+                 */
+                function getTextWidth(input) {
+                    // re-use canvas object for better performance
+                    const canvas = getTextWidth.canvas || (getTextWidth.canvas = document.createElement('canvas'));
+                    const context = canvas.getContext('2d');
+                    context.font = '14px Roboto';
+
+                    return context.measureText(input).width;
+                }
 
                 /**
                  * Table initialization callback. This will hide the loading indicator.
--- a/src/content/styles/modules/_filters.scss
+++ b/src/content/styles/modules/_filters.scss
@@ -16,6 +16,10 @@
             height: 100%;
             min-height: 0; // Firefox fix; otherwise div won't shrink: http://stackoverflow.com/questions/27424831/firefox-flexbox-overflow
 
+            table {
+                table-layout: fixed; // make autowidth: false works for datatable https://datatables.net/forums/discussion/30530/set-fixed-column-width
+            }
+
             .dataTables_wrapper {
                 height: 100%;
                 display: flex !important; // there was an update in datatables styles that was overriding `flex` with `block`
@@ -35,6 +39,34 @@
                         background: lighten($divider-color, 20%);
 
                         tbody {
+
+                            // ellipsis renderer when text is longer then field width
+                            .rv-render-ellipsis {
+                                width: 100%;
+                                border: none;
+                                background: none;
+                                padding: 0;
+                                font-size: 14px;
+                                cursor: text;
+                                overflow: hidden;
+                                text-overflow: ellipsis;
+                            }
+
+                            .rv-render-tooltip {
+                                display: none;
+                            }
+
+                            // on keyboard focus show tooltip to be wcag compliant
+                            .rv-render-ellipsis:focus + .rv-render-tooltip {
+                                display: block;
+                                position: absolute;
+                                border: 1px solid #D3D3D3;
+                                background-color: #F2F2F2;
+                                padding: 2px 5px;
+                                box-shadow: 1px 1px 2px #CCC;
+                                max-width: 250px;
+                                white-space: normal;
+                            }
 
                             .rv-data {
                                 // flex: flex-grow flex-shrink flex-basis; IE sets flex-basis as 0px which collapses the span instead of 0% as Chrome does;


AleksueiR:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/app/ui/filters/filters-default.directive.js, line 259 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-K_5A1VzleybLQcsyVyL:-K_5A1VzleybLQcsyVyM:b-kzts23) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/bb5633d875ec59b722c1734344ba0899a4334c38/src/app/ui/filters/filters-default.directive.js#L259)):*
> ```JavaScript
> 
>                 /**
>                  * Render long text width ellipsis (https://datatables.net/blog/2016-02-26)
> ```

Long string can be rendered with ellipsis using CSS only (even though it looks truncated on the page, double click and Ctl+C will actually copy the full string). I don't think using a rendered for that provides much benefit here. 

---

*[src/app/ui/filters/filters-default.directive.js, line 292 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-K_5DjubNZtArhp2LZs8:-K_5DjubNZtArhp2LZs9:b-ofp3fm) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/bb5633d875ec59b722c1734344ba0899a4334c38/src/app/ui/filters/filters-default.directive.js#L292)):*
> ```JavaScript
> 
>                         // for wcag we add a text input read only. This element is focusable so we can have tooltips.
>                         return `<input type="text" readonly title="${esc(text)}" value="${esc(text)}"
> ```

Adding readonly input fields to the overflowing cells is an interesting idea, but I see several problems:

Fields are not added to all the cells and any keyboard-only user will probably be confused why only come cells in the table are tabable.

With the mouse you get a double tooltip when hovering over: 


Our current implementation of the focus manager tries to keep the focus on the element, preventing the table from scrolling (in the gif I'm trying to scroll down, but it keeps resetting as soon as the focused row scrolls away):

We probably need to tweak the focus manager to avoid this behaviour, but until then, we need another solution for this jumping focus.

@fgpv-vpgf/project-members Ideas anyone?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546)*

jolevesq:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/app/ui/filters/filters-default.directive.js, line 259 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-K_5A1VzleybLQcsyVyL:-K_Zwrk7vnLwRxxydiVo:b8q8389) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/bb5633d875ec59b722c1734344ba0899a4334c38/src/app/ui/filters/filters-default.directive.js#L259)):*

Long string can be rendered with ellipsis using CSS only (even though it looks truncated on the page, double click and Ctl+C will actually copy the full string). I don't think using a rendered for that provides much benefit here. 
</blockquote></details>

I know, the renderellipsis function is use to render long string for keyboard user. 

---

*[src/app/ui/filters/filters-default.directive.js, line 292 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-K_5DjubNZtArhp2LZs8:-K_Zx853vMqQpPOKBocL:bhab7qv) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/bb5633d875ec59b722c1734344ba0899a4334c38/src/app/ui/filters/filters-default.directive.js#L292)):*

Adding readonly input fields to the overflowing cells is an interesting idea, but I see several problems:

Fields are not added to all the cells and any keyboard-only user will probably be confused why only come cells in the table are tabable.

With the mouse you get a double tooltip when hovering over: 


Our current implementation of the focus manager tries to keep the focus on the element, preventing the table from scrolling (in the gif I'm trying to scroll down, but it keeps resetting as soon as the focused row scrolls away):

We probably need to tweak the focus manager to avoid this behaviour, but until then, we need another solution for this jumping focus.

@fgpv-vpgf/project-members Ideas anyone?
</blockquote></details>

I think for the first problem user will understand quickly why it doesn't stop at all cells. If we need we can put a tabbable element in all cells.

I don't really think the double tooltip will be a problem because user who will use keyboard to navigate the table will surely not use a mouse. Navigate the table with the keyboard is very time consuming and it is hard to find the information quickly. We can do something to avoid double tooltip. Same for the focus (the problem occurs on the zoom and details button as well)... The focus, if we scroll down a little bit faster, is removed and it works. Maybe we just need to replicate this even if the scroll is very slow.

Do we have a use case for a user who will use mouse and keyboard? 

We have this implementation in GCViz and we never received a comment about it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546)*

AleksueiR:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/app/ui/filters/filters-default.directive.js, line 259 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-K_5A1VzleybLQcsyVyL:-K__0A7u1jSN2bEMhNqo:bgtxygu) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/bb5633d875ec59b722c1734344ba0899a4334c38/src/app/ui/filters/filters-default.directive.js#L259)):*
<details><summary><i>Previously, jolevesq (Johann Levesque) wrote…</i></summary><blockquote>

I know, the renderellipsis function is use to render long string for keyboard user. 
</blockquote></details>

No sure I understand how it helps keyboard users. 
CSS ellipsis can be removed when the cell gets focus, so the full string is visible.

---

*[src/app/ui/filters/filters-default.directive.js, line 292 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-K_5DjubNZtArhp2LZs8:-K__0dNz2xOH0TMASTLt:b4zikv4) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/bb5633d875ec59b722c1734344ba0899a4334c38/src/app/ui/filters/filters-default.directive.js#L292)):*
<details><summary><i>Previously, jolevesq (Johann Levesque) wrote…</i></summary><blockquote>

I think for the first problem user will understand quickly why it doesn't stop at all cells. If we need we can put a tabbable element in all cells.

I don't really think the double tooltip will be a problem because user who will use keyboard to navigate the table will surely not use a mouse. Navigate the table with the keyboard is very time consuming and it is hard to find the information quickly. We can do something to avoid double tooltip. Same for the focus (the problem occurs on the zoom and details button as well)... The focus, if we scroll down a little bit faster, is removed and it works. Maybe we just need to replicate this even if the scroll is very slow.

Do we have a use case for a user who will use mouse and keyboard? 

We have this implementation in GCViz and we never received a comment about it.
</blockquote></details>

I agree, they might understand what and why is happening, but it is still unpredictable - you can't tell how far the focus will jump if the next truncated field is hidden from view (it even might jump to a different row if the currently focused cell is the last truncated field). I also not sure how this will work with screen readers trying to read the table. As far as I know, arrow keys are used to navigate tables using screen readers, not the Tab key. Does the cell get focus when a screen reader reads it? If so, then the tooltip will show up and the screen reader will say it as well.

The second tooltip is created by the `title` attribute on the `input` field. Do we need it? It's not strictly correct anyway, since it's not the title of the field. Maybe we can remove this and avoid the double tooltip issue.

We don't have a specific use case to simultaneous mouse/keyboard use, but we try not to punish users if they want to switch from one to the other midway. 

If you scroll a little bit faster, the focus is removed, and since the cell which had focus is already removed from the dom, the focus manager __will__ put the focus somewhere else (previously focused item, as far as I understand).
Here's what was happening for me - after scrolling faster, the focus was set to the header cell, but since that header cell was outside of the view, it was scrolled into the view, which messed up column alignment (scrolling some more fixed it though).


This one is focus manager's fault and we need to address it somehow (not in this PR though).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546)*

jolevesq:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/app/ui/filters/filters-default.directive.js, line 259 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-K_5A1VzleybLQcsyVyL:-K__7G1e_9rT8BWYjLA6:b-chb9q2) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/bb5633d875ec59b722c1734344ba0899a4334c38/src/app/ui/filters/filters-default.directive.js#L259)):*

No sure I understand how it helps keyboard users. 
CSS ellipsis can be removed when the cell gets focus, so the full string is visible.
</blockquote></details>

cells does not get focus by default. It is why I created a readonly input when cell string are too long. If we only use css, keyboard user can't focus the cell and get the info. I can remove the render function but we will have to do something for the keyboard user.

---

*[src/app/ui/filters/filters-default.directive.js, line 292 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-K_5DjubNZtArhp2LZs8:-K__8RZDTNdEbrD3nLXO:b-58r00u) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/bb5633d875ec59b722c1734344ba0899a4334c38/src/app/ui/filters/filters-default.directive.js#L292)):*

I agree, they might understand what and why is happening, but it is still unpredictable - you can't tell how far the focus will jump if the next truncated field is hidden from view (it even might jump to a different row if the currently focused cell is the last truncated field). I also not sure how this will work with screen readers trying to read the table. As far as I know, arrow keys are used to navigate tables using screen readers, not the Tab key. Does the cell get focus when a screen reader reads it? If so, then the tooltip will show up and the screen reader will say it as well.

The second tooltip is created by the `title` attribute on the `input` field. Do we need it? It's not strictly correct anyway, since it's not the title of the field. Maybe we can remove this and avoid the double tooltip issue.

We don't have a specific use case to simultaneous mouse/keyboard use, but we try not to punish users if they want to switch from one to the other midway. 

If you scroll a little bit faster, the focus is removed, and since the cell which had focus is already removed from the dom, the focus manager __will__ put the focus somewhere else (previously focused item, as far as I understand).
Here's what was happening for me - after scrolling faster, the focus was set to the header cell, but since that header cell was outside of the view, it was scrolled into the view, which messed up column alignment (scrolling some more fixed it though).


This one is focus manager's fault and we need to address it somehow (not in this PR though).
</blockquote></details>

The datatable can't be navigate out of the box with the arrow. The arrow scroll horizontaly. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546)*

jolevesq:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/app/ui/filters/filters-default.directive.js, line 259 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-K_5A1VzleybLQcsyVyL:-K__ATiFJhll9yhZzMI1:b-ijmd81) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/bb5633d875ec59b722c1734344ba0899a4334c38/src/app/ui/filters/filters-default.directive.js#L259)):*
<details><summary><i>Previously, jolevesq (Johann Levesque) wrote…</i></summary><blockquote>

cells does not get focus by default. It is why I created a readonly input when cell string are too long. If we only use css, keyboard user can't focus the cell and get the info. I can remove the render function but we will have to do something for the keyboard user.
</blockquote></details>

the render ellipsis function is use to set the class for textoverfloew only when needed. This way I can set a class:hover css to show the tooltip on hover

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546)*

AleksueiR:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/app/ui/filters/filters-default.directive.js, line 259 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-K_5A1VzleybLQcsyVyL:-K__AdCMcN1raG-svx51:b5sfzng) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/bb5633d875ec59b722c1734344ba0899a4334c38/src/app/ui/filters/filters-default.directive.js#L259)):*
<details><summary><i>Previously, jolevesq (Johann Levesque) wrote…</i></summary><blockquote>

the render ellipsis function is use to set the class for textoverfloew only when needed. This way I can set a class:hover css to show the tooltip on hover
</blockquote></details>

We can keep the input field and use CSS to truncate with ellipsis, no?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546)*

AleksueiR:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/app/ui/filters/filters-default.directive.js, line 292 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-K_5DjubNZtArhp2LZs8:-K__BAjd3X3RM-auBD2S:b-7w1oix) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/bb5633d875ec59b722c1734344ba0899a4334c38/src/app/ui/filters/filters-default.directive.js#L292)):*
<details><summary><i>Previously, jolevesq (Johann Levesque) wrote…</i></summary><blockquote>

The datatable can't be navigate out of the box with the arrow. The arrow scroll horizontaly. 
</blockquote></details>

I think screen readers have a mode or something to navigate table. Without a screen reader, I'm not sure what the best way to navigate a table.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546)*

jolevesq:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/app/ui/filters/filters-default.directive.js, line 259 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-K_5A1VzleybLQcsyVyL:-K__BJB1DVWLzPVGQKHv:b-4y3lzr) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/bb5633d875ec59b722c1734344ba0899a4334c38/src/app/ui/filters/filters-default.directive.js#L259)):*

We can keep the input field and use CSS to truncate with ellipsis, no?
</blockquote></details>

The input field is generated by the renderEllipsis function associated on column.render of datatable. The ellipsis is done by CSS, the only thing this function does is to replace td by the input if the text is too long. We can use the row callback function to replace td by the input when we create the table.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546)*

jolevesq:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/app/ui/filters/filters-default.directive.js, line 292 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-K_5DjubNZtArhp2LZs8:-K__DPWVTeW-IWyBYiPR:b-gibi56) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/bb5633d875ec59b722c1734344ba0899a4334c38/src/app/ui/filters/filters-default.directive.js#L292)):*

Great, thanks for explaining. Got confused here what this acctually does :)
</blockquote></details>

Can you resume me what should I do for this PR.

Do I remove input field and use text-overflow on the td element? Do you want me to use the row creation callback instead of renderrer (this can be use if we want to apply the same thing for all the row)? I need to set title equal to the content if I want it to be visible on hover even if you said it should be the field title. I can set a td:hover to see the tooltip and let title to be the field name but we will end with 2 tooltips.

Thanks

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546)*

jolevesq:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/app/ui/filters/filters-default.directive.js, line 292 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-K_5DjubNZtArhp2LZs8:-K__Ousxdbqad6Z6ZLdI:b47xnbq) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/bb5633d875ec59b722c1734344ba0899a4334c38/src/app/ui/filters/filters-default.directive.js#L292)):*
<details><summary><i>Previously, jolevesq (Johann Levesque) wrote…</i></summary><blockquote>

Can you resume me what should I do for this PR.

Do I remove input field and use text-overflow on the td element? Do you want me to use the row creation callback instead of renderrer (this can be use if we want to apply the same thing for all the row)? I need to set title equal to the content if I want it to be visible on hover even if you said it should be the field title. I can set a td:hover to see the tooltip and let title to be the field name but we will end with 2 tooltips.

Thanks
</blockquote></details>

I found an extension (KeyTable) to navigate the table with the arrow. There is also event (on focus and on blur) we can use to show text if renderer with ellipsis.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546)*

jolevesq:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/app/ui/filters/filters-default.directive.js, line 292 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-K_5DjubNZtArhp2LZs8:-K__Qbe7Vr6KJO5RB7V6:b-czzvsv) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/bb5633d875ec59b722c1734344ba0899a4334c38/src/app/ui/filters/filters-default.directive.js#L292)):*
<details><summary><i>Previously, jolevesq (Johann Levesque) wrote…</i></summary><blockquote>

I found an extension (KeyTable) to navigate the table with the arrow. There is also event (on focus and on blur) we can use to show text if renderer with ellipsis.
</blockquote></details>

Take a look at https://datatables.net/extensions/keytable/examples/initialisation/scroller.html There is a bug with the scroller extension (not able to got to the last item). Can we use this and report the bug to KeyTable extension or we need to solve it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546)*

AleksueiR:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/app/ui/filters/filters-default.directive.js, line 259 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-K_5A1VzleybLQcsyVyL:-K__FX7VORiGuOpp3XKL:b-8fx2u7) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/bb5633d875ec59b722c1734344ba0899a4334c38/src/app/ui/filters/filters-default.directive.js#L259)):*
<details><summary><i>Previously, jolevesq (Johann Levesque) wrote…</i></summary><blockquote>

The input field is generated by the renderEllipsis function associated on column.render of datatable. The ellipsis is done by CSS, the only thing this function does is to replace td by the input if the text is too long. We can use the row callback function to replace td by the input when we create the table.
</blockquote></details>

:+1: 

---

*[src/app/ui/filters/filters-default.directive.js, line 292 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-K_5DjubNZtArhp2LZs8:-K__Zya2B9BXYhViMU_Z:b-t8bgmw) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/bb5633d875ec59b722c1734344ba0899a4334c38/src/app/ui/filters/filters-default.directive.js#L292)):*
<details><summary><i>Previously, jolevesq (Johann Levesque) wrote…</i></summary><blockquote>

Take a look at https://datatables.net/extensions/keytable/examples/initialisation/scroller.html There is a bug with the scroller extension (not able to got to the last item). Can we use this and report the bug to KeyTable extension or we need to solve it?
</blockquote></details>

Not sure what is going on with titles. When I try to remove them, that custom tooltip shows up on focus, but not standard browser tooltip. I don't think it's a big problem anyway.

I suggest we merge this and make new issue to address the focus manager problem with jumping focus, to include the table navigation, and (if possible) address that extension's problem with not being able to access the last record (should report that bug to KeyTable extension as well). Does this sound okay?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546)*

jolevesq:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/app/ui/filters/filters-default.directive.js, line 292 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-K_5DjubNZtArhp2LZs8:-K__beT0YHmvMXYOy1xm:b6wn0wg) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/bb5633d875ec59b722c1734344ba0899a4334c38/src/app/ui/filters/filters-default.directive.js#L292)):*

Not sure what is going on with titles. When I try to remove them, that custom tooltip shows up on focus, but not standard browser tooltip. I don't think it's a big problem anyway.

I suggest we merge this and make new issue to address the focus manager problem with jumping focus, to include the table navigation, and (if possible) address that extension's problem with not being able to access the last record (should report that bug to KeyTable extension as well). Does this sound okay?
</blockquote></details>

Ok, I have already put table extension inside this PR to test. Do you want me to remove it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546)*

AleksueiR:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/app/ui/filters/filters-default.directive.js, line 292 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-K_5DjubNZtArhp2LZs8:-K__dfrDTdYeXM1ptjAB:by1qguv) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/bb5633d875ec59b722c1734344ba0899a4334c38/src/app/ui/filters/filters-default.directive.js#L292)):*
<details><summary><i>Previously, jolevesq (Johann Levesque) wrote…</i></summary><blockquote>

Ok, I have already put table extension inside this PR to test. Do you want me to remove it?
</blockquote></details>

Does it work? Can't test at the moment as we are having issues with our firewall :( Keep it in, if it works, we can tweak it later.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546)*

jolevesq:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/app/ui/filters/filters-default.directive.js, line 292 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-K_5DjubNZtArhp2LZs8:-K__hZ0ghUlOxPg_Jdsg:bpn8wpc) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/bb5633d875ec59b722c1734344ba0899a4334c38/src/app/ui/filters/filters-default.directive.js#L292)):*

Does it work? Can't test at the moment as we are having issues with our firewall :( Keep it in, if it works, we can tweak it later.
</blockquote></details>

It works, I can put the options false at the moment and we can re-enable it to solve the issue with the scroller

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546)*

AleksueiR:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/app/ui/filters/filters-default.directive.js, line 292 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-K_5DjubNZtArhp2LZs8:-K__hyKXjDryRaxuj1HW:bbjq473) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/bb5633d875ec59b722c1734344ba0899a4334c38/src/app/ui/filters/filters-default.directive.js#L292)):*
<details><summary><i>Previously, jolevesq (Johann Levesque) wrote…</i></summary><blockquote>

It works, I can put the options false at the moment and we can re-enable it to solve the issue with the scroller
</blockquote></details>

Good idea, please do that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546)*

jolevesq:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/app/ui/filters/filters-default.directive.js, line 292 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-K_5DjubNZtArhp2LZs8:-K_a3pPVqRsprZpFxUUC:bb6un1j) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/bb5633d875ec59b722c1734344ba0899a4334c38/src/app/ui/filters/filters-default.directive.js#L292)):*

Good idea, please do that.
</blockquote></details>

Done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546)*

milesap:
Works on iPhone and mobile layouts :+1: 

---

Reviewed 1 of 3 files at r2, 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1546#-:-K_ewfPpUrDwQaLL9cjF:b-97612h)*

PR #1580, opened 2017-01-12
fix(ui): added rv-ignore-focusout to focus manager

### LINKED ISSUE #1557: Focus manager resets lost focus at inopportune moments
The following comments are from https://github.com/fgpv-vpgf/fgpv-vpgf/pull/1546 PR.

Our current implementation of the focus manager tries to keep the focus on the element, preventing the table from scrolling (in the gif I'm trying to scroll down, but it keeps resetting as soon as the focused row scrolls away):

We probably need to tweak the focus manager to avoid this behaviour, but until then, we need another solution for this jumping focus.

If you scroll a little bit faster, the focus is removed, and since the cell which had focus is already removed from the dom, the focus manager __will__ put the focus somewhere else (previously focused item, as far as I understand).
Here's what was happening for me - after scrolling faster, the focus was set to the header cell, but since that header cell was outside of the view, it was scrolled into the view, which messed up column alignment (scrolling some more fixed it though).

milesap:
## Description
Any child of an element with the rv-ignore-focusout property is ignored
when it loses focus, and the next tab press is handled by the browser.

Closes #1557

## Testing
:eye: 

## Documentation
inline where needed

## Checklist

- [x] code passes unit tests

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/1580)

### DIFF
--- a/src/app/focus-manager.js
+++ b/src/app/focus-manager.js
@@ -29,6 +29,8 @@
     let lockFocus = false;
     // used to call cancelTimeout during focus if focusout timeout has started
     let focusoutTimerCancel;
+    // when true focus manager will only consider history elements for focus movement
+    let restoreFromHistory = false;
 
     /**
      * Represents one viewer on a page, with multiple viewers being possible. Tracks viewer state,
@@ -365,6 +367,7 @@
      * @param  {Object} event the keydown event object
      */
     function onKeydown(event) {
+        /*jshint maxcomplexity:9 */
         const viewerActive = viewerGroup.status(statuses.ACTIVE);
         const viewerWaiting = viewerGroup.status(statuses.WAITING);
         keys[event.which] = true;
@@ -375,8 +378,9 @@
                 viewerActive.setStatus(statuses.INACTIVE);
             // shiftFocus must return true indicating focus has been moved, and only then
             // do we want to prevent the browser from moving focus itself
-            } else if (event.which === 9 && shiftFocus(!event.shiftKey)) { // tab keydown only
+            } else if (event.which === 9 && shiftFocus(!event.shiftKey, restoreFromHistory)) { // tab keydown only
                 event.preventDefault(true);
+                restoreFromHistory = false;
             }
 
         } else if (viewerWaiting) {
@@ -411,7 +415,10 @@
      */
     function onFocusout(event) {
         const viewer = viewerGroup.status(statuses.ACTIVE);
-        if (viewer && !lockFocus && event.relatedTarget === null) {
+        if ($(event.target).closest('[rv-ignore-focusout]').length > 0) {
+            restoreFromHistory = true;
+
+        } else if (viewer && !lockFocus && event.relatedTarget === null) {
             // Allow for a short time as determined by focusoutDelay in milliseconds so that when focus
             // leaves unexpectedly, focus can be manually set and we don't need to be back through history
             // Animations often cause focus loss when, for example, one element is being hidden while the
--- a/src/app/ui/filters/filters-default.html
+++ b/src/app/ui/filters/filters-default.html
@@ -5,7 +5,7 @@
     title-value="{{ 'filter.title' | translate }}{{ self.display.requester.name }}"
     is-loading="self.display.isLoading">
 
-    <div class="rv-filters">
+    <div class="rv-filters" rv-ignore-focusout>
 
         <!--md-button ng-click="self.draw()">click this button if you don't see the table below</md-button-->
 


AleksueiR:
This will let the focus escape from the viewer to the host page permanently.
What about my suggestion here: https://github.com/fgpv-vpgf/fgpv-vpgf/issues/1557#issuecomment-271391734 ? Is it much harder to implement? 

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1580#-:-KaIX13YLsga-IvuNR58:b9k48uw)*

milesap:
It "escapes" but only briefly. By allowing the browser to handle the next focus movement it remains inside the viewer. 

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1580#-:-KaIXgT9jkr--89v7Lix:b-aqd8he)*

AleksueiR:
(Using keyboard) on `fgp-en` page, when the focus is lost in the datatable, the table stops scrolling (the page scrolls instead - using arrow keys), pressing `Tab` does not move focus anywhere at all; pressing `Shift + Tab` resets the focus to the viewer, not the datatable.

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1580#-:-KaIYb79A02nC3Jk9Blx:baqlh40)*

milesap:
Fixed focus from leaving when you scroll with keyboard. Scrolling however will stop since focus is no longer inside the scrollable area. 

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1580#-:-KaIa16p6gLjok-_v8Qr:b-hysq06)*

alyec:
Reviewed 1 of 2 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1580)*

AleksueiR:
:lgtm:

---

Reviewed 1 of 2 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1580#-:-KaJCAtpT0gt-f-aDKz9:bnfp4nl)*

jolevesq:
> index-one.html
In Chrome, if I keep my finger on down arrow to scroll down after few seconds it stop scrolling. If I press tab it brings me back where I started to scroll. Shift-Tab focus on the last column header.

---

Reviewed 1 of 2 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1580#-:-KaN1PB_NvGz1Vg0D_ui:b154nck)*

milesap:
Yeah same for me - this is ok since it is default browser behavior. We could consider improving this functionality (new issue?) where arrow keys continuously scroll the datatable. 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1580#-:-KaN2UdkSmmAWnCkmC5K:b-az3ho5)*

jolevesq:
Ok, maybe what I am working with the keytable extension will solve it :lgtm: 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1580#-:-KaN3271MwK3XFeBUK3m:b-j3rn7c)*

jolevesq:
Is it possible to completely exclude a node from the focus manager? It would solve the problem with datatables.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1580#-:-KagcMQShga7ziGp_rdO:baubefw)*

milesap:
I don't recommend this for now as many unwanted side effects could be introduced. There are many ways to manipulate the focus manager from it's default behavior. Along with `rv-ignore-focusout` there is also `rv-trap-focus`. If these don't work you can also manually link elements together with the `link` method. Have a look at them and let me know if you have any questions. If you can't get it to work as expected open an issue and I'll take a look.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1580#-:-Kagceuaa23yHq171S4f:b-karlw1)*

milesap:
Pushed change where on tab key press, focus is restored from history only. This can be tested in datatables. Note that scrolling too far destroys the datatable row, so focus moves to the header. From there focus moves to the first row in the datatable which has a rendered element. If you only scroll a little the row is not destroyed and it receives focus.

---

Review status: 1 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1580#-:-KahHNTIVyoN7mM4_JZw:b-tukg59)*

alyec:
Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1580)*

AleksueiR:
:lgtm:

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1580#-:-KarFAdr0zHVFDxg-8BU:bnfp4nl)*

jolevesq:
:lgtm:

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1580#-:-KarFS3_NDBvgsjAnaxK:bnfp4nl)*

PR #1734, opened 2017-03-02
fix(focus): improve export dialog support with focus manager

### LINKED ISSUE #1623: Keyboard focus on export dialog window
There is some issues with the focus on the export dialog window.

- On open, the focus should go on the close button
- When custom size is selected from the drop down menu, focus should go to width setting (the first input inside custom size setting)
- When we tab from width to height, there is a hidden focus
     - If we remove the label element it solves the focus problem
     - Focus seems to go back to the window. If we press the arrow key, the window will scroll.
     - Disabling the focus manager solve the problem. Focus manager seems to do something with the label element inside the md-input-container

milesap:
## Description
Note that the following point in the issue

> On close with esc, the focus should go back to export menu item and side panel should reopen

is not addressed here as this is more related to side-nav and menu interaction. The same behaviour is seen with the help dialog. We can discuss and open a separate issue if need be for such a change.

Closes #1623

## Testing
🙄 

## Documentation
inline, jsdocs

## Checklist

- [x] code passes unit tests

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/1734)

### DIFF
--- a/src/app/focus-manager.js
+++ b/src/app/focus-manager.js
@@ -12,7 +12,7 @@
     // jQuery selectors for elements that are likely focusable
     // since we don't want focus on md-sidenav (focus instead on close button) we omit this with [tabindex]:not(md-sidenav)
     const focusSelector = `a[href], area[href], input:not([disabled]), select:not([disabled]), textarea:not([disabled]),
-        button:not([disabled]), iframe, object, embed, [tabindex]:not(md-sidenav), [contenteditable]`;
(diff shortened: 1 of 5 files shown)


jolevesq:
Reviewed 5 of 5 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1734)*

milesap:
Implementation will be combined with #1623 solution to be PR'd shortly.

Issue #1373, opened 2016-11-15
Add arrow key support for md-menu


milesap:
While `md-menu` does have built-in arrow key support, it is disabled by the focus manager. Down arrow is equivalent to tab, and up arrow is equivalent to shift + tab.

milesap:
This issue is going to take a bit more time to get right. So I don't forget over the weekend:

- Angular material is manhandling events, with a vengeance for the down arrow key. On down keypress focus is always reset to the top of the menu. 
  See: https://github.com/angular/material/blob/34b331040c0c9cf0e8c329838cb1083b285ad050/src/components/autocomplete/autocomplete.spec.js#L54
Also:
https://github.com/angular/material/search?utf8=%E2%9C%93&q=isNumPadKey

I have a working solution if I use another key in place of the down arrow. Up arrow is not affected surprisingly. Took a while to realize it wasn't FM...

TODO: Disable angular from bypassing event filters.

PR #1741, opened 2017-03-07
feat(focus): Add arrow support for md-menu and other improvements

### LINKED ISSUE #1623: Keyboard focus on export dialog window

milesap:
## Description
- Removed requirement for calling init(), now initialized before
app-seed but after jQuery (required)
- Added new HTML and jQuery element prototype rvFocus() which replaces focus(true) when
setting focus on viewer elements. They are now exclusive and will raise
a warning when used incorrectly.
- Added attribute rv-focus-init which allows the use of focus() on a
viewer element or one of its children if focus is outside the element
- Added jQuery element prototype nextFocus() which moves focus to the
next element
- Documentation added

Closes #1623, #1373

## Testing
Yes, quite a bit

## Documentation
Of course!

## Checklist

- [x] code passes unit tests

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/1741)

### DIFF
--- a/docs/developer/technical/focus_manager.md
+++ b/docs/developer/technical/focus_manager.md
@@ -2,6 +2,16 @@
 
 To be accessible and WCAG 2 compliant, the viewer needs to support keyboard users as well as mice-wielding ones (by extension, dual-class power users who use both devices also need to be thought of). The focus manager is therefore responsible for allowing users to navigate through the viewer seamlessly, making content easy to reach no matter what input device is used.
 
+### Element Attributes
+
+| Name | Description |
+| **** | *********** |
+| rv-trap-focus         |  Once focus enters an element or child of this element, focus movement is restricted within it. If focus is at the end of the trap, it is moved to the first focusable element in the trap. Focus can only leave if the focus manager is disabled, or the entire trapped element is not focusabled (i.e. is hidden, destroyed). The value must be the viewers HTML id. This can be used on elements whos HTML is rendered outside the viewers HTML so that focus manager includes it as part of the viewers focus movement flow.
+| rv-focus-status       |  Added to the main viewer HTML element with a value of either NONE, INACTIVE, WAITING, ACTIVE to signify the current focus state of the viewer. This is useful for outside libraries to see the current state of any given viewer.
+| rv-ignore-focusout    |  Normally the focus manager attempts to recover from a loss of focus by moving back through its history. This attribute disables this behaviour so no recovery is initiated.
+| rv-focus-init         |  Allows the use of focus() on the viewer element with this attribute or any of its children if focus is not current on the element or any of its children. This allows outside libraries to set initial focus normally, yet restricts them on subsequent movement attempts.
+
+
 #### Important concepts you should know
 
 Before we go into how it works, there are some key concepts you'll need to be aware of that affect all parts of the viewer:
@@ -9,19 +19,21 @@ Before we go into how it works, there are some key concepts you'll need to be aw
 - The `$rootElement` has an `rv-trap-focus` property set in `bootstrap.js` (more on this later)
 - Any element which is not a descendent of `$rootElement` but should be considered a part of the viewer must have an `rv-trap-focus` property set with a value equal to the `$rootElement` id property
 - The javascript prototype functions `focus`, `preventDefault`, `stopPropagation`, and `stopImmediatePropagation` are disabled for all elements and descendents of an `rv-trap-focus`.
-    - To set focus, use `focusService.setFocus(myElement)`
-    - `preventDefault`, `stopPropagation`, and `stopImmediatePropagation` can be used by passing an extra parameter `true` as in `$('#myElement').preventDefault(true)`. You should never prevent or stop the bubbling of `mousedown`,`keydown` or `keyup` as they are needed for the focus manager to work effectively.
-- The tabindex of all focusable elements inside the viewer (except for the map) is set to -1. You should not set or rely on tabindex for proper tab order.
+    - To set focus, use `element.rvFocus()`
+    - `preventDefault`, `stopPropagation`, and `stopImmediatePropagation` can be used by passing an extra parameter `true` as in `element.preventDefault(true)`. You should never prevent or stop the bubbling of `mousedown`,`keydown` or `keyup` as they are needed for the focus manager to work effectively.
+- The tabindex of all browser focusable elements inside a viewer is set to -1. You should not set or rely on tabindex for proper tab order.
+- A tabindex of -2 is a special value which indicates that the viewer element is focusable.
+- A tabindex of -3 is a special value which indicates that the viewer element is not only focusable, it is focusable through the focus() prototype property which is usually restricted. This is useful for granting outside libraries focus access to a specific element.
 
 #### rv-trap-focus
 
-An element with this property ensures that focus will not leave, unless the element itself is destroyed or hidden. When the last element is reached and the `tab` key is pressed the first focusable element is focused. 
+An element with this property ensures that focus will not leave, unless the element itself is destroyed or hidden. When the last element is reached and the `tab` key is pressed the first focusable element is focused.
 
 #### How it works
 
 When focus reaches the viewer in an embedded page, a dialog appears asking if the user would like to enter the viewer, or provides the option to skip over it. In full screen viewers, or when navigated to by mouse, the viewer is activated automatically. If the user chooses to skip the viewer, the tab key moves focus to the next element outside the viewer. Since all viewer elements have a tabindex of `-1`, this is handled natively by the browser itself. If the user chooses to enter the viewer, the focus manager intercepts all tab key presses and handles the focus movement.
 
-It's common for focused elements to be destroyed or hidden during normal viewer use such as when a panel is closed or a menu disappears. When this happens, the last known focusable element is automatically focused. 
+It's common for focused elements to be destroyed or hidden during normal viewer use such as when a panel is closed or a menu disappears. When this happens, the last known focusable element is automatically focused.
 
 To escape the viewer the user must press `escape` + `tab`. This disables focus management and the tab key naturally moves to the next focusable element outside the viewer. Mouse clicking outside the viewer also disables it.
 
@@ -30,4 +42,4 @@ To escape the viewer the user must press `escape` + `tab`. This disables focus m
 There are two instances you should use the focus manager:
 
 - When a user action (like a click or keypress) changes the viewer and you'd like focus moved to some other area of the viewer. A common scenario is when a dialog opens and you'd like focus moved to the close button.
-- When you want to override the default focus movement behaviour between two elements. This is done automatically in table of contents when you select `settings` and the settings panel opens. A link between the toc layer and the setting panel has been created. 
+- When you want to override the default focus movement behaviour between two elements. This is done automatically in table of contents when you select `settings` and the settings panel opens. A link between the toc layer and the setting panel has been created.
--- a/gulp.config.js
+++ b/gulp.config.js
@@ -51,6 +51,7 @@ module.exports = function () {
             'global-registry.js',
             'app.js',
             'templates.js',
+            'focus-manager.js',
             'app-seed.js',
             'corePlugins.js',
         ],
--- a/src/app/bootstrap.js
+++ b/src/app/bootstrap.js
@@ -313,7 +313,6 @@
     // load core.js last and execute any deferred polyfills/patches
     loadScript(`${repo}/core.js`, () => {
         RV._deferredPolyfills.forEach(dp => dp());
-        RV.focusManager.init();
         RV.allScriptsLoaded = true;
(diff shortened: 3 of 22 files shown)


AleksueiR:
Review status: 0 of 18 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/app/ui/export/custom-size.directive.js, line 21 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1741#-Kehvfp--UJGBYUeVW7Y:-Kehvfp--UJGBYUeVW7Z:b-6oh956) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/2c1d42dcae45602542611dafde9a26e158c60d28/src/app/ui/export/custom-size.directive.js#L21)):*
> ```JavaScript
>             restrict: 'E',
>             templateUrl: 'app/ui/export/custom-size.html',
>             link: (scope, el) => el.find('input').first().focus(true)
> ```

Should this be `rvFocus`?

---

*[src/app/ui/export/custom-size.html, line 1 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1741#-Kehtra1UVrJPFAqqgWk:-Kehtra2xzJdsP3vBQIA:b-ym6i52) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/2c1d42dcae45602542611dafde9a26e158c60d28/src/app/ui/export/custom-size.html#L1)):*
> ```HTML
> <md-input-container ng-disabled="self.isError">
> ```

Why do you need to split out this template section? There should be a different way to move the focus without splitting this into a separate directive.

The focus is moved to the "width" field when the export dialog itself is opened, not when the custom size sub-panel is opened.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1741)*

milesap:
Review status: 0 of 18 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/app/ui/export/custom-size.directive.js, line 21 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1741#-Kehvfp--UJGBYUeVW7Y:-KehxXNADK33Dy9cMr5l:b-3p9coh) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/2c1d42dcae45602542611dafde9a26e158c60d28/src/app/ui/export/custom-size.directive.js#L21)):*

Should this be `rvFocus`?
</blockquote></details>

Missed this one, changed it

---

*[src/app/ui/export/custom-size.html, line 1 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1741#-Kehtra1UVrJPFAqqgWk:-KehxpD225SD1slOjHu2:buy7j4b) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/2c1d42dcae45602542611dafde9a26e158c60d28/src/app/ui/export/custom-size.html#L1)):*

Why do you need to split out this template section? There should be a different way to move the focus without splitting this into a separate directive.

The focus is moved to the "width" field when the export dialog itself is opened, not when the custom size sub-panel is opened.
</blockquote></details>

There is an `md-content` parent element with `ng-if="self.exportSizes.isCustomOptionSelected()"` so that content is destroyed and re-created each time custom size option is selected in the dropdown. This just sets focus each time it is created (the "angular way" I suppose, instead of watching the `ng-if` for changes and setting focus that way. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1741)*

jolevesq:
I found few issues but the good point is I am able to navigate with tab in FF and Safari

When we close with esc, the focus didn't go back to export inside the menu panel. Focus only goes to the button to open the panel.
There is a hidden focus between the size selection and the setting button.
When I tab to select custom size and go back to size selector with tab and select Default the focus goes to url bar. It should stay on size selector.

I know it is not part of this PR but if you select Share then press esc, focus goes on the map not the Share button. Same with Basemap, when we close it focus don't go tot he right place. When we change language, panel close and focus goes  to the wrong place and when I tab and I am on menu, it stay white (don't see the selection). A new issue should be created to deal with focus and sidenav panel.

---

Reviewed 16 of 18 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/app/ui/filters/filters-default.directive.js, line 415 at r2](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1741#-KeiIOGROihx_4nO7C6_:-KeiIOGROihx_4nO7C6a:btamx4w) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/e4b2700c19af65a5fde294d0a4566b8f4cc69b42/src/app/ui/filters/filters-default.directive.js#L415)):*
> ```JavaScript
>                     // at the same time it solve a problem because when focus is on menu button, even if focus is on the
>                     // table cell it goes inside the menu and loop through it at the same time as we navigate the table
>                     $timeout(() => { $rootElement.find('[type=\'filters\'] button.rv-close').rvFocus(); }, 100);
> ```

Whit the rv-close-button, do we still need this? should it go automatically to the close button?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1741#-:-KeiDB5lXXkOtSm-OuZI:bqa2b9e)*

jolevesq:
In FF and Chrome when I select something in the menu show/hide all, the focus doesn't stay on the icon to open the menu. It goes at the en of TOC for Safari and it is lost then goes to map with FF. Works in Chrome.

When panel is close, focus doesn't go to export button. In Chrome it goes to menu, in FF and Safari it is lost and the first tab goes to overview map

hidden tab solved!
Change from custom size solved!

---

Review status: 14 of 19 files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1741#-:-KejEMO1dTvOdlHXdBV6:b-2mk2ot)*

AleksueiR:
Reviewed 13 of 18 files at r1, 1 of 2 files at r2, 5 of 5 files at r3.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1741)*

milesap:
Thanks for the detailed testing! Please only test in chrome for now, there is no FF or IE support. If any part of FM works in those browsers then bonus points :+1: 

Made some fairly large breaking changes, so a good run through with chrome would be a good idea.

---

Review status: 15 of 21 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/app/ui/filters/filters-default.directive.js, line 415 at r2](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1741#-KeiIOGROihx_4nO7C6_:-KejSgXxMDahu5tBuOH3:bks9wx7) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/e4b2700c19af65a5fde294d0a4566b8f4cc69b42/src/app/ui/filters/filters-default.directive.js#L415)):*
<details><summary><i>Previously, jolevesq (Johann Levesque) wrote…</i></summary><blockquote>

Whit the rv-close-button, do we still need this? should it go automatically to the close button?
</blockquote></details>

Good point, I haven't had time to test this out but yes it should work (and I'll update this on a subsequent FM update, this one is getting large)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1741#-:-KejTCmkRx-_pyO0ExKw:b-17049c)*

jolevesq:
I made a good run an list some issues (only from Chrome this time :smile: ). I don't know if they are all part of this PR but it can help for future WCAG/focus tasks

If I open a menu like collapse all and instead of the arrow to navigate I press tab, the focus goes to url bar and menu can't close. For menu when there is more then one active option you can tab and it will react like press arrow. If there is only one choice, press tab will set focus outside the menu.

when press zoom to location it shows a toast message. The focus should go to the close button of this message instead of having to tab. When we close the message the focus is lost when it should go back to zoom to location.

When we open the help it doesn't focus to close button. When click on close it goes back to help button :thumbsup: 

When press enter on geosearch focus doesn't go to the text input of geosearch. We have to tab once.

Load file or service when I press close button, focus goes back to add layer  :thumbsup: but when I press esc, the TOC is close and focus goes to layer button when it should go on add layer

Datatable seems good. I am able to open from button and panel and when i close from esc or button focus goes at the right place

When open Metadata from the layer setting, focus should go to close button. When we close something open from an item inside a menu I am not sure if focus should go back to the menu item or the button to open the menu is ok?

When we close the export, focus should go back to export button inside sidenav. 

Just a tought, when we tab from map focus it should go to menu instead of overview map. Overview should be focus after  menu panel. This way the focus don't go right/left/right and menu is more important to go to then overview map.

---

Reviewed 2 of 5 files at r3, 5 of 6 files at r4.
Review status: 20 of 21 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1741#-:-KejXcR4UDQ8-6VRB-9J:b71e8yx)*

milesap:
> If I open a menu like collapse all and instead of the arrow to navigate I press tab, the focus goes to url bar and menu can't close. For menu when there is more then one active option you can tab and it will react like press arrow. If there is only one choice, press tab will set focus outside the menu.

Fixed

> When we open the help it doesn't focus to close button. When click on close it goes back to help button 👍

Fixed

> When press enter on geosearch focus doesn't go to the text input of geosearch. We have to tab once.

Not fixed in this PR. To implement this there are two options:
- Set focus on the input with rvFocus
- Set input tabindex to -3 and let autocomplete set focus itself
This would need a decorator as the input is not available in `geosearch-bar.js` link function

> Load file or service when I press close button, focus goes back to add layer 👍 but when I press esc, the TOC is close and focus goes to layer button when it should go on add layer

This is correct behaviour for FM since the TOC is closed, layer button is only valid target.

> When open Metadata from the layer setting, focus should go to close button. When we close something open from an item inside a menu I am not sure if focus should go back to the menu item or the button to open the menu is ok?

This is correct behaviour

> When we close the export, focus should go back to export button inside sidenav.

This is correct behaviour for FM since the sidenav is closed on export open, menu button is only valid target.

> Just a tought, when we tab from map focus it should go to menu instead of overview map. Overview should be focus after menu panel. This way the focus don't go right/left/right and menu is more important to go to then overview map.

I like this idea. Perhaps in the next update this will be done, but to keep this PR from touching any more files I'll skip this for now. Overview map is inside `<div class="rv-esri-map" rv-init-map tabindex="0" translate translate-attr-aria-label="geo.aria.mapdescription"></div>` I believe so it would need some poking around to change its ordering.

---

Review status: 18 of 22 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1741#-:-KeskKygwkQg28kXJ2TF:b-1rp117)*

jolevesq:
Ok to put geosearch in another issue. It will have to be solve because hidden focus is a WCAG fail.

Maybe it can be part of another issue as well but the esc on export bring back to layer instead of add layer is a little bit confusing. We should discuss on this.

We just miss the auto-focus on close when metadata and settings opens and focus is not trap inside the panel.

This is a good FM progress :thumbsup: 

---

Reviewed 1 of 6 files at r4, 4 of 4 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1741#-:-KfHVdEb6pTYf7q3RMhq:b-rawri0)*

jolevesq:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1741#-:-KfHXzLi2BbklHnQhGuF:bnfp4nl)*

milesap:
ty, opened #1770 for geosearch. The others will be under consideration for future implementation.

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1741#-:-KfH__v-1YEMojQ4NUAU:b4njmq)*

PR #1844, opened 2017-04-10
fix(angular): allow multiple versions of angular on a page

### LINKED ISSUE #960: Handle cases where hosting page has Angular already
From Pauline George (NACEI):

> We tried to embed the map viewer and it is returning some errors.  After some investigation the errors maybe due to conflicts with the core.js and bootstrap.js files.  It seems like the conflict occurs if a site or application is already using angular (we are using angular).  There seems to be an open issue in github for this particular issue

milesap:
## Description
This allows for multiple angular versions to run on the same page. This still requires additional testing however it does currently work if:
- host page has angular app initialized prior to our, or;
- loads their version of angular after our application loads and scopes it into their bootstrap

Closes #960

## Testing
some, more needed, just an idea

## Documentation
inline

## Checklist

- [x] code passes unit tests

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/1844)

### DIFF
--- a/gulp.config.js
+++ b/gulp.config.js
@@ -40,7 +40,8 @@ module.exports = function () {
             src + 'index-jso.html',
             src + 'index-fgp-en.html',
             src + 'index-fgp-fr.html',
-            src + 'bookmark-decode.html'
+            src + 'bookmark-decode.html',
+            src + 'multiple-angular.html'
         ],
         indexProtractor: src + 'index-protractor.html',
 
--- a/src/app/app-seed.js
+++ b/src/app/app-seed.js
@@ -1,28 +1,32 @@
-/**
- * @function app.core.seed
- * @inner
- * @desc `seed` is an implicit function which runs on application startup to
- * initialize all marked DOM nodes to map instances
- */
-angular.element(document)
-    .ready(() => {
-        'use strict';
-        // NOTE: let and const cannot be used in this file due to protractor problems
+$(function () {
+    (function (angular) {
+        /**
+         * @function app.core.seed
+         * @inner
+         * @desc `seed` is an implicit function which runs on application startup to
+         * initialize all marked DOM nodes to map instances
+         */
+        angular.element(document)
+            .ready(() => {
+                'use strict';
+                // NOTE: let and const cannot be used in this file due to protractor problems
 
-        // The app nodes in the dom
-        window.RV._nodes.forEach(node => {
+                // The app nodes in the dom
+                window.RV._nodes.forEach(node => {
 
-            // load shell template into the node
-            // we need to create an explicit child under app's root node, otherwise animation
-            // doesnt' work; see this plunk: http://plnkr.co/edit/7EIM71IOwC8h1HdguIdD
-            // or this one: http://plnkr.co/edit/Ds8e8d?p=preview
-            node.appendChild(angular.element('<rv-shell class="md-body-1">')[0]);
+                    // load shell template into the node
+                    // we need to create an explicit child under app's root node, otherwise animation
+                    // doesnt' work; see this plunk: http://plnkr.co/edit/7EIM71IOwC8h1HdguIdD
+                    // or this one: http://plnkr.co/edit/Ds8e8d?p=preview
+                    node.appendChild(angular.element('<rv-shell class="md-body-1">')[0]);
 
-            // bootstrap each node as an Angular app
-            // strictDi enforces explicit dependency names on each component: ngAnnotate should find most automatically
-            // this checks for any failures; to fix a problem add 'ngInject'; to the function preamble
-            angular.bootstrap(node, ['app'], {
-                strictDi: true
+                    // bootstrap each node as an Angular app
+                    // strictDi enforces explicit dependency names on each component: ngAnnotate should find most automatically
+                    // this checks for any failures; to fix a problem add 'ngInject'; to the function preamble
+                    angular.bootstrap(node, ['app'], {
+                        strictDi: true
+                    });
+                });
             });
-        });
-    });
+    })(angular);
+});
--- a/src/app/bootstrap.js
+++ b/src/app/bootstrap.js
@@ -51,7 +51,7 @@
             get url() { return `//cdn.datatables.net/${this.ourVersion}/js/jquery.dataTables.min.js`; }
         }
     };
-    const dependenciesOrder = ['jQuery', 'dataTables', 'angular'];
+    const dependenciesOrder = ['jQuery', 'dataTables'];
 
     const d = document;
     const scripts = d.getElementsByTagName('script'); // get scripts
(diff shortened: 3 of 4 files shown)


milesap:
Added a working demo page for running multiple angular versions on a single page. Our viewer now loads its own version if an identical version is not already loaded on host. 

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1844#-:-KhX39K5E_rHNQvGfreL:b-gyx7p6)*

alyec:
Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1844)*

alyec:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/app/bootstrap.js, line 98 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1844#-KiV3QW8zoJDigntEzGi:-KiV3QW8zoJDigntEzGj:bvlgohy) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/aaa1db4ab2ab3207d308ba1015e74ae0c616af15/src/app/bootstrap.js#L98)):*
> ```JavaScript
>     // host page can continue to use their version if their app is loaded before ours, or they can manually
>     // bootstrap a scoped angular object like we do in app-seed.js
>     if (window.angular && window.angular.version.full !== dependencies.angular.ourVersion) {
> ```

Alex and I had a bit of discussion about this.  We can probably try to save the reference to the original Angular loaded on the page here and restore it after the callback in `loadScript('core.js' ...` .  This would essentially do something like `.noConflict()` for angular while still allowing all our `angular-*` libraries to load via `core.js`.  Should be fairly easy to test and if it works we aren't destroying the global scope or resorting to iframes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1844)*

AleksueiR:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/app/bootstrap.js, line 98 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1844#-KiV3QW8zoJDigntEzGi:-KiVEyuFYSuYMRe99nk6:boihdt0) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/aaa1db4ab2ab3207d308ba1015e74ae0c616af15/src/app/bootstrap.js#L98)):*
<details><summary><i>Previously, alyec (Aly Merchant) wrote…</i></summary><blockquote>

Alex and I had a bit of discussion about this.  We can probably try to save the reference to the original Angular loaded on the page here and restore it after the callback in `loadScript('core.js' ...` .  This would essentially do something like `.noConflict()` for angular while still allowing all our `angular-*` libraries to load via `core.js`.  Should be fairly easy to test and if it works we aren't destroying the global scope or resorting to iframes.
</blockquote></details>

See here for an example: http://www.mattburkedev.com/multiple-angular-versions-on-the-same-page/

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1844)*

milesap:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/app/bootstrap.js, line 98 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1844#-KiV3QW8zoJDigntEzGi:-Kj3AQkoIM-U3zRXsFJ6:b-reqods) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/aaa1db4ab2ab3207d308ba1015e74ae0c616af15/src/app/bootstrap.js#L98)):*

See here for an example: http://www.mattburkedev.com/multiple-angular-versions-on-the-same-page/
</blockquote></details>

Angular will not load if `window.angular` already contains a version of angular. This can be bypassed by setting `window.angular` to an empty object or null. 

Restoring the original angular version on `loadScript(`${repo}/core.js`, () => {` will cause errors for anything which is not executed immediately (as angular will then refer to the original angular version rather than the viewers version). This could be corrected by not referencing `angular` in our app but something like `RV.angular` which contains our verstion. This would affect every file however.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1844)*

AleksueiR:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/app/bootstrap.js, line 98 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1844#-KiV3QW8zoJDigntEzGi:-Kj3BziXdG9UkUmC8hyK:bi6v241) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/aaa1db4ab2ab3207d308ba1015e74ae0c616af15/src/app/bootstrap.js#L98)):*
<details><summary><i>Previously, milesap (Miles Petrov) wrote…</i></summary><blockquote>

Angular will not load if `window.angular` already contains a version of angular. This can be bypassed by setting `window.angular` to an empty object or null. 

Restoring the original angular version on `loadScript(`${repo}/core.js`, () => {` will cause errors for anything which is not executed immediately (as angular will then refer to the original angular version rather than the viewers version). This could be corrected by not referencing `angular` in our app but something like `RV.angular` which contains our verstion. This would affect every file however.
</blockquote></details>

Sounds like iframes...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1844)*

alyec:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/app/bootstrap.js, line 98 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1844#-KiV3QW8zoJDigntEzGi:-Kj3BzUfjRGZdm4JPLMN:b-l9h4gh) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/aaa1db4ab2ab3207d308ba1015e74ae0c616af15/src/app/bootstrap.js#L98)):*

Sounds like iframes...
</blockquote></details>

Yeah, that would be needed, but that's also why we have IIFEs everywhere

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1844)*

milesap:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/app/bootstrap.js, line 98 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1844#-KiV3QW8zoJDigntEzGi:-Kj3I7nsq5udUy1ZI82G:bdjzsr) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/aaa1db4ab2ab3207d308ba1015e74ae0c616af15/src/app/bootstrap.js#L98)):*
<details><summary><i>Previously, alyec (Aly Merchant) wrote…</i></summary><blockquote>

Yeah, that would be needed, but that's also why we have IIFEs everywhere
</blockquote></details>

Tests within `shell.directive.js` confirm the angular version is the pages version, not our own. Maybe we'll have to leave this for now and hope webpack saves the day :pray: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1844)*

milesap:
Not ready for release, webpack should be able to handle this well.

Issue #1839, opened 2017-04-06
fix(bug): missing menu icon


jolevesq:
Some menu icons (visibility all and add layers) are missing since last rebase

<img width="659" alt="screen shot 2017-04-06 at 8 24 05 am" src="https://cloud.githubusercontent.com/assets/3472990/24757030/f0b931fc-1aab-11e7-9f09-06029580e99e.png">

AleksueiR:
Took a quick look and it seems something is stripping the icons after their templates have been compiled. Very random.

milesap:
Ok'd by @mweech to cancel v1.6.2 and move straight to v1.7.0

PR #1846, opened 2017-04-13
draft: icons as a font

milesap:
Got carried away with #1839, somehow I ended up experimenting with icons as fonts instead of svg. This PR should not be merged. 

- Updated `angular-translate` to v2.15.1 which fixes an issue where `translate` needed to be added for `translate-attr-aria-label` to work. This might be opened as a separate issue since `translate` tries to translate the entire contents of the element. This was causing issues with md-icon inside a button with `translate`

- AM allows icons in the form `<md-icon>iconname</md-icon>`. Prefixes such as `map` or `action` are not needed.

- Most file chanes are just `md-icon` formatting. `icon.decorator.js` and `bootstrap.js` contain the changes.

- This would eliminate `makesvgcache` - but we would still require a static svg file(s) to be injected into `core.js` for the cases where we want to use custom svg images.

We most likely won't want to change the way we handle icons this drastically, but I figured I'd show my work none the less. Enjoy!

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/1846)

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -18,7 +18,7 @@
     "angular-material": "1.1.3",
     "angular-minicolors": "^0.0.9",
     "angular-sanitize": "1.4.12",
-    "angular-translate": "^2.13.0",
+    "angular-translate": "^2.15.1",
     "angular-translate-loader-static-files": "^2.13.0",
     "bezier-easing": "^1.0.0",
     "canvas-toBlob.js": "https://github.com/eligrey/canvas-toBlob.js.git",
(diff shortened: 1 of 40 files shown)


AleksueiR:
Can you split this into two separate PRs? Icons and translate fixes?

---

Review status: 0 of 40 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/1846#-:-KhbUe-LWs8vF1A12rzD:b-xmx2zk)*

PR #1891, opened 2017-05-18
Moving hovertip template

AleksueiR:
## Description
- move hovertip template to tootlip service as the default template
- prevent controlled layers from being queried (even if specified in the config)
- prevent controlled layers from having enabled bounding boxes (even if specified in the config)
- prevent controlled layers from having hovertips

## Testing
Nope.

## Documentation
Inline comments.

## Checklist

- ~~[ ] commits messages follow the guidelines~~ nope
- [x] code passes unit tests
- ~~[ ] release notes have been updated~~ not yet

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/1891)


AleksueiR:
Re-targeted at develop.

PR #2418, opened 2017-10-06
Ability to toggle ranges via legend

### LINKED ISSUE #2368: Ability to toggle ranges via legend
image

As previously discussed for CESI. The "EcoGeo" way to do this was to add the geodatabase to the service several times for each data range or symbol, and filter the geodatabase on those parameters. However, this makes the data grid impractical for CESI, so we need to develop a method for toggling ranges in the mapping interface. These selections should also toggle what is visible in the datagrid, however this is lesser priority. 

Sample service: http://section917.cloudapp.net/arcgis/rest/services/CESI/CESI_Air_Ambient_OzoneAverage/MapServer

Here are some Acceptance Criteria (by @mweech):
1. Esri layer renderers should allow visibility to be toggled.  This will allow for layers to be filtered via their individual symbology stack.
2. The viewer should not maintain visibility state of the symbology stack if the layer itself has it's visibility change.
3. Parent layer visibility will always override renderer visibility.
4. Renderer visibility should persist when using full state bookmarks
5. Datagrid should reflect the visibility of the renderers
5. Autolegend output will by default allow for the toggling of renderer visibility
6. Structured legend should allow for the toggling of renderer visibility based on the following values being present in the config:
       a. the custom symbol config element has a mapping to a specific layer renderer
       b. the custom symbol config element specifies renderer visibility as toggleable
       c. the custom symbol config element can optionally specify an initial state (default: visible)
7. Print map legend output will not omit hidden layer renderers from the legend, however the map itself will reflect the current visibility state.

WMS layer toggling will be considered out of scope for this issue.

milesap:
## Description
Allows symbology to be toggled. New controls added when symbology is expanded. Symbology is filtered both on the map and table (if open).

Can be disabled in config under layer config as `toggleSymbology: false`. It's true by default.

Do not merge until James gives the :white_check_mark: since geoApi is being beefed up to handle more edge cases on the `definitionClause`.

@dan-bowerman got you on this as well to make sure this works for CESI.

Closes #2368

## Testing
46 `Dynamic Layer with only one child with filter -> Power Plants".  You can also test sample 1 crops but dt is very slow.

## Documentation
inline

## Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/2418)

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "datatables.net-select": "1.2.2",
     "dotjem-angular-tree": "github:dotJEM/angular-tree",
     "file-saver": "1.3.3",
-    "geoApi": "github:fgpv-vpgf/geoApi#v2.1.0-19",
+    "geoApi": "github:fgpv-vpgf/geoApi#v2.1.0-20",
     "gsap": "1.20.2",
     "jquery": "2.2.4",
     "jquery-hoverintent": "1.8.2",
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "datatables.net-select": "1.2.2",
     "dotjem-angular-tree": "github:dotJEM/angular-tree",
     "file-saver": "1.3.3",
-    "geoApi": "github:fgpv-vpgf/geoApi#develop",
+    "geoApi": "github:fgpv-vpgf/geoApi#v2.1.0-20",
     "gsap": "1.20.2",
     "jquery": "2.2.4",
     "jquery-hoverintent": "1.8.2",
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "datatables.net-select": "1.2.2",
     "dotjem-angular-tree": "github:dotJEM/angular-tree",
     "file-saver": "1.3.3",
-    "geoApi": "github:fgpv-vpgf/geoApi#v2.1.0-20",
+    "geoApi": "github:fgpv-vpgf/geoApi#develop",
     "gsap": "1.20.2",
     "jquery": "2.2.4",
     "jquery-hoverintent": "1.8.2",
--- a/schema.json
+++ b/schema.json
@@ -529,6 +529,7 @@
                 "name": { "type": "string", "description": "The display name of the layer.  If it is not present the viewer will make an attempt to scrape this information." },
                 "url": { "type": "string", "description": "The service endpoint of the layer.  It should match the type provided in layerType." },
                 "layerType": { "type": "string", "enum": [ "esriDynamic", "esriFeature", "esriImage", "esriTile", "ogcWms" ] },
+                "toggleSymbology": { "type": "boolean", "default": true, "description": "Allows individual symbols to have visibility toggled on/off."},
                 "extent": { "$ref": "#/definitions/extentWithReferenceNode" }
             },
             "required": [ "id", "layerType", "url" ]
@@ -560,6 +561,7 @@
                 "metadataUrl": { "type": "string", "default": null, "description": "The metadata url of the layer service" },
                 "catalogueUrl": { "type": "string", "default": null, "description": "The catalogue url of the layer service" },
                 "layerType": { "type": "string", "enum": [ "esriFeature" ] },
+                "toggleSymbology": { "type": "boolean", "default": true, "description": "Allows individual symbols to have visibility toggled on/off."},
                 "tolerance": { "type": "number", "default": 5, "description": "Specifies the tolerance in pixels when determining if a feature was clicked. Should be non-negative integer" },
                 "extent": { "$ref": "#/definitions/extentWithReferenceNode" },
                 "controls": { "$ref": "#/definitions/legendEntryControls" },
@@ -603,6 +605,7 @@
                 "metadataUrl": { "type": "string", "default": null, "description": "The metadata url of the layer service" },
                 "catalogueUrl": { "type": "string", "default": null, "description": "The catalogue url of the layer service" },
                 "layerType": { "type": "string", "enum": [ "esriDynamic" ] },
+                "toggleSymbology": { "type": "boolean", "default": true, "description": "Allows individual symbols to have visibility toggled on/off."},
                 "singleEntryCollapse": { "type": "boolean", "default": false, "description": "Indicates that the dynamic layer with a single layer entry should be rendered without the root group." },
                 "layerEntries": {
                     "type": "array",
--- a/src/app/core/config.class.js
+++ b/src/app/core/config.class.js
@@ -547,6 +547,7 @@ function ConfigObjectFactory(Geo, gapiService, common) {
             this._disabledControls = defaultedSource.disabledControls;
             this._userDisabledControls = defaultedSource.userDisabledControls;
             this._initialFilteredQuery = defaultedSource.initialFilteredQuery;
+            this._toggleSymbology = typeof source.toggleSymbology === 'boolean' ? source.toggleSymbology : true;
 
             // remove metadata control if no metadata url is specified after applying defaults
             if (!source.metadataUrl) {
@@ -569,6 +570,7 @@ function ConfigObjectFactory(Geo, gapiService, common) {
 
         get initialFilteredQuery() { return this._initialFilteredQuery; }
         set initialFilteredQuery(value) { this._initialFilteredQuery = value; }
+        get toggleSymbology() { return this._toggleSymbology; }
 
         _hovertipEnabled = false;
         get hovertipEnabled () {        return this._hovertipEnabled; }
--- a/src/app/core/constant.service.js
+++ b/src/app/core/constant.service.js
@@ -66,7 +66,9 @@ function events($rootScope) {
         rvHighlightDetailsItem: 'rvHighlightDetailsItem',
         rvGeosearchClose: 'rvGeosearchClose', // Fire when geosearch close
         rvTableReady: 'rvTableReady',
-        rvFeatureMouseOver: 'rvFeatureMouseOver'
+        rvFeatureMouseOver: 'rvFeatureMouseOver',
+
+        rvLayerDefinitionClauseChanged: 'rvLayerDefinitionClauseChanged'
     };
 }
 
--- a/src/app/focus-manager.js
+++ b/src/app/focus-manager.js
@@ -20,7 +20,7 @@ const focusSelector = [
     'md-switch:not([disabled])',
     'md-slider:not([disabled])',
     'textarea:not([disabled])',
-    'button:not([disabled])',
+    'button:not([disabled]):not([nofocus])',
     '.rv-esri-map',
     '[tabindex=-2]',
     '[tabindex=-3]'
@@ -331,7 +331,8 @@ function elemIsFocusable(index, element) {
         el.css('opacity') !== 0 &&
         // avoid setting focus on closing menu items
         !el.parents().hasClass('md-leave') &&
-        !el.parents().hasClass('md-leave-add');
+        !el.parents().hasClass('md-leave-add') &&
+        !el.is('[nofocus]');
 }
 
 /**
--- a/src/app/ui/table/table.service.js
+++ b/src/app/ui/table/table.service.js
@@ -91,7 +91,7 @@ function tableService(stateManager, geoService, $rootScope, $q, gapiService, deb
      */
     function init() {
 
-        events.$on('definitionClauseChgd', () => {
+        events.$on(events.rvLayerDefinitionClauseChanged, () => {
             if (filterTimeStamps.onCreated !== null) {
                 filteredState().then(() => {
                     filterTimeStamps.onChanged = Date.now();
--- a/src/app/ui/table/table.service.js
+++ b/src/app/ui/table/table.service.js
@@ -12,7 +12,7 @@ angular
     .factory('tableService', tableService);
 
 function tableService(stateManager, geoService, $rootScope, $q, gapiService, debounceService, $rootElement, $timeout,
-    referenceService, layerRegistry, configService, Geo) {
+    referenceService, layerRegistry, configService, Geo, events) {
 
     // timestamps can be watched for key changes to filter data
     const filterTimeStamps = {
@@ -90,6 +90,15 @@ function tableService(stateManager, geoService, $rootScope, $q, gapiService, deb
      * @private
      */
     function init() {
+
+        events.$on('definitionClauseChgd', () => {
+            if (filterTimeStamps.onCreated !== null) {
+                filteredState().then(() => {
+                    filterTimeStamps.onChanged = Date.now();
+                });
+            }
+        });
+
         $rootScope.$on('extentChange', debounceService.registerDebounce(onExtentChange, 300, false));
 
         // DataTable is either being created or destroyed
@@ -633,27 +642,21 @@ function tableService(stateManager, geoService, $rootScope, $q, gapiService, deb
     }
 
     /**
-     * Determines the type of filters to apply and sets validOIDs.
+     * Applies validOIDs returned from queryMap or all oids if no query is applied.
      *
      * @function filteredState
      * @private
      * @return  {Promise}   resolves to undefined when the filtering is complete
      */
     function filteredState() {
-        return $q(resolve => {
-            if (service.filter.isActive) {
-                queryMapserver().then(oids => {
-                    validOIDs = oids;
-                    resolve();
-                });
-            } else {
-                // convert existing rows into a validOIDs list (no filtering applied)
-                validOIDs = stateManager.display.table.data.rows.map(
-                    row => parseInt(row[stateManager.display.table.data.oidField])
-                );
-                resolve();
-            }
-        });
+        const layerRecId = layerRegistry.getLayerRecord(stateManager.display.table.requester.legendEntry.layerRecordId);
+
+        // we're not filtering by either symbology or extent, so resolve with all oid's as valid
+        if (!service.filter.isActive && typeof layerRecId.definitionClause !== 'string') {
+            return $q.resolve(stateManager.display.table.data.rows.map(row => parseInt(row[stateManager.display.table.data.oidField])));
+        }
+
+        return queryMapserver().then(oids => (validOIDs = oids));
     }
 
     /**
@@ -669,10 +672,13 @@ function tableService(stateManager, geoService, $rootScope, $q, gapiService, deb
         const legEntry = state.requester.legendEntry;
         const layerRecId = layerRegistry.getLayerRecord(legEntry.layerRecordId);
 
-        const queryOpts = {
-            geometry: geoService.map.extent,
-            outFields: [state.data.oidField]
-        };
+        const filterByExtent = service.filter.isActive;
+        const filterBySymbology = typeof layerRecId.definitionClause === 'string';
+
+        const queryOpts = { outFields: [state.data.oidField] };
+        if (filterByExtent) {
+            queryOpts.geometry = geoService.map.extent;
+        }
 
         // query the layer itself instead of making a mapserver request
         if (legEntry.parentLayerType === 'esriFeature' && legEntry.layerType === 'esriFeature' && layerRecId.isFileLayer()) {
@@ -683,7 +689,18 @@ function tableService(stateManager, geoService, $rootScope, $q, gapiService, deb
 
         // only include oidField values after previous mapserver query resulted in a exceededTransferLimit exception
         if (lastOID > 0) {
-            queryOpts.where = `${state.data.oidField} > ${lastOID}`;
+            queryOpts.where = `(${state.data.oidField} > ${lastOID})`;
+
+            if (filterBySymbology) {
+                queryOpts.where += ` AND `;
+            }
+        } else {
+            queryOpts.where = ''; // this is needed for our string concatenation below, else its undefined
+        }
+
+
+        if (filterBySymbology) {
+            queryOpts.where += `(${layerRecId.definitionClause})`;
         }
 
         return gapiService.gapi.query.queryGeometry(queryOpts).then(featureSet => {
--- a/src/app/ui/table/table.service.js
+++ b/src/app/ui/table/table.service.js
@@ -653,7 +653,8 @@ function tableService(stateManager, geoService, $rootScope, $q, gapiService, deb
 
         // we're not filtering by either symbology or extent, so resolve with all oid's as valid
         if (!service.filter.isActive && typeof layerRecId.definitionClause !== 'string') {
-            return $q.resolve(stateManager.display.table.data.rows.map(row => parseInt(row[stateManager.display.table.data.oidField])));
+            validOIDs = stateManager.display.table.data.rows.map(row => parseInt(row[stateManager.display.table.data.oidField]));
+            return $q.resolve();
         }
 
         return queryMapserver().then(oids => (validOIDs = oids));
--- a/src/app/ui/toc/symbology-stack.directive.js
+++ b/src/app/ui/toc/symbology-stack.directive.js
@@ -9,6 +9,21 @@ const RV_DESCRIPTION_ITEM = '.rv-description-container';
 const RV_DURATION = 0.3;
 const RV_SWIFT_IN_OUT_EASE = Power1;
 
+class ToggleSymbol {
+    constructor(block, name) {
+        this.block = block;
+        this.nameField = this.block.mainProxyWrapper._proxy._source.nameField;
+        this.name = name;
+        this.isSelected = this.block.visibility;
+    }
+
+    click() { this.isSelected = !this.isSelected; }
+
+    get icon() { return this.isSelected ? "toggle:check_box" : "toggle:check_box_outline_blank"; }
+
+    get query() { return this.isSelected ? `${this.nameField}='${this.name}'` : null; }
+}
+
 /**
  * @module rvSymbologyStack
  * @memberof app.ui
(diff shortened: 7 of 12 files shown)


james-rae:
Some test services for various cases (still trying to find a default symbol renderer one)

1-field unique (warning, service acting flakey, maybe use another)
http://maps-cartes.ec.gc.ca/arcgis/rest/services/CESI/CESI_AirEmissions_CO/MapServer/0

class break
http://section917.cloudapp.net/arcgis/rest/services/TestData/ClassBreaks/MapServer/0

two field unique (with renderer -> legend aggregation)
http://geoappext.nrcan.gc.ca/arcgis/rest/services/NEB/Incident/MapServer/0

tile (no filters)
http://maps-cartes.ec.gc.ca/arcgis/rest/services/Basemaps/OilSandsBasemap/MapServer

simple (no filters)
http://section917.cloudapp.net/arcgis/rest/services/TestData/BasicCity/MapServer/0

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-Kvmqur1GscTRizDbKqB:b9tz1rl)*

dan-bowerman:
The English rebuilt CESI services are all ready to roll as well: 
http://section917.cloudapp.net/arcgis/rest/services/CESI

So far, so good on my end! Looks and functions great!

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KvmrWhbUj6MvtXdaEu_:b-ex2uvy)*

JahidAhmed:
All the datatables are empty, is there any reason for that?

---

Review status: 0 of 9 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-Kvms9h1uN3GeO7zBqBy:b9gn1n8)*

dan-bowerman:
WOMM? Only the toggled on data is displayed. This sample only works with the Power Plants service. 


---







*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KvmtKzp5NDuq_N53anj:b-fe1avs)*

JahidAhmed:
Oh yep, that's working as expected but I wasn't too sure if the other datatables were empty by design somehow? (Don't know too much about this PR)

---

Review status: 0 of 9 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KvmtnOswP3zRoq3RmTl:b-7tqc1y)*

milesap:
I'll fix, some are empty but shouldn't be

---

Review status: 0 of 9 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-Kvmu361GoVYsZf7jOEX:b3h36vt)*

JahidAhmed:
Also, when you toggle the entire layer to be off, should the symbology also be toggled off or is it supposed to stay on?

---

Review status: 0 of 9 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KvmuahYlgUAkL6t-5Mk:b5is4tk)*

milesap:
My vote is for the symbology to remain as is rather then reset them but that can be up for debate. I'm wondering if their initial visibility should follow initial layer visibility as it does now or always be on/off.

---

Review status: 0 of 9 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KvmvPWhNwumK_oJ1ryN:b-jfvo0f)*

james-rae:
If we can have a grid open on an invisible layer, then I think we should not tie the checkboxes to layer visibility.  Else the grid will clear if you set the visibility off

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-Kvmzabsk21m7bmMKN1h:b-d5cqa)*

dan-bowerman:
^ Especially for CESI's Protected Areas layer, which for performance reasons will be a tile cache, but we want to put an invisible layer "on top" for identifying/grid purposes.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KvmzzpPRXk2kMjeOEcd:b-uzadbm)*

james-rae:
Efficiency - When all checkboxes are checked, we are having the app evaluate all the logic for every symbol. Would be better if we can detect ALL-ON state and essentially clear the filter from the grid and layer. Especially relevant if grid is open (as we end up asking the server for all the ids, then tell the grid “hey, all these are good”).

UI - if layer is invisible, and we touch a symbol checkbox, it will turn back on the visibility. Probably should de-couple this. Was also able to get visibility controls on main legend and layer settings panel to go out-of-synch. Can also get a state where all checkboxes are turned off, but then start toggling visibility and see all points re-appear on the map.

Grid - if grid is open and all symbol filters are unchecked, the grid shows all rows. Likely caused because all-off layers results in an empty string for the definition query. The logic treats this as “no definition” and acts accordingly (show all rows).

UI - I get no checkboxes when loading a class breaks layer. Try with [Sea Parks](http://section917.cloudapp.net/arcgis/rest/services/TestData/ClassBreaks/MapServer/0).  Same with this [two-field unique value layer](http://geoappext.nrcan.gc.ca/arcgis/rest/services/NEB/Incident/MapServer/0) in feature layer mode (they appear in dynamic mode).

Efficiency - It’s looking like with feature layers it’s re-querying all geometry, which is unfortunate but is controlled by ESRI API. The network traffic looks really inefficient; I believe it’s doing queries by bounding boxes across the map, but I can see the same features coming down in multiples (so overlap). Noticeable redraw for each click.  If we deem this unacceptable, a workaround would be to essentially run the grid filter logic, and use the result array of ids to turn on/off the visibility of the actual on-client features. We no longer set the definition query on the feature layer object. Less network traffic on symbol filter change, but possibly more on every pan/zoom (would need some analysis). Needs more code to manage the feature visibility and ensure things get reapplied on pan/zoom (especially if on-demand layer, which is always re-creating on-client features).

---

Reviewed 8 of 9 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/app/ui/toc/templates/symbology-stack.html, line 14 at r2](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-KvnEP6OwOn3RnVyY_jC:-KvnEP6OwOn3RnVyY_jD:b7n3x90) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/7342df2a5b5c86bd18e61370efcdaa3fb4615b38/src/app/ui/toc/templates/symbology-stack.html#L14)):*
> ```HTML
>             <span class="rv-symbol-label">{{ symbol.name }}</span>
>         </div>
>         
> ```

evil spaces

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KvnQFGIdbeMV0bTXIT4:b-neavz8)*

AleksueiR:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/app/ui/toc/templates/symbology-stack.html, line 14 at r2](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-KvnEP6OwOn3RnVyY_jC:-Kw59uLg6m7gRercDOU8:b-rzaq68) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/7342df2a5b5c86bd18e61370efcdaa3fb4615b38/src/app/ui/toc/templates/symbology-stack.html#L14)):*

evil spaces
</blockquote></details>

To battle evil spaces, add this to your vscode settings:


---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418)*

milesap:
- When all symbols are checked, the query is removed instead of having a query for all symbols
- Changing symbology visibility has no effect on layer visibility
- symbology visibility is always on by default
- Fixed issue where datatable showed results when all symbols were deselected
- The Sea Parks layer now shows the symbolofy toggle boxes, however the actual toggling is a hit or a miss. It first throws a lot of errors, then after a few tries reloads and then works. 

---

Review status: 7 of 9 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-Kw5lKWK2Hnggx2HQf6r:b-ll3cbf)*

AleksueiR:
Nice, works pretty well - I like how you can toggle parts of the `Plant Hardiness` layer :+1: 

___

Errors are thrown when the details panel is opened:
> angular.js:14642 TypeError: Cannot read property 'layerRecordId' of undefined
    at Object.l [as link] (symbology-stack.directive.js:104)
    at angular.js:1385
    at angular.js:10514
    at Ae (angular.js:10520)


#### Checkboxes are badly aligned in some cases (looks like it depends on the layer's nestedness level):

###### not enough


###### too far


#### Sometimes checkboxes won't show up
No checbox during the opening animation here, but it will appear when closing:





---

Reviewed 4 of 9 files at r1.
Review status: 7 of 9 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/app/ui/toc/symbology-stack.directive.js, line 132 at r3](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-KwAJ9rK4jZxiN5sW8qL:-KwAJ9rK4jZxiN5sW8qM:btx4sar) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/5c6e3677bded78f9a61f0877e7b01dc28f44f27f/src/app/ui/toc/symbology-stack.directive.js#L132)):*
> ```JavaScript
> 
>             // trigger event which table uses to update
>             events.$broadcast('definitionClauseChgd');
> ```

Can you add this event to the events list to avoid hardcoding random strings?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwAJsx8VNo7Z_uljcBy:b-57pqm9)*

AleksueiR:
Reviewed 2 of 9 files at r1, 1 of 2 files at r3.
Review status: 8 of 9 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418)*

milesap:
Should be all fixed up :+1: 

---

Review status: 7 of 10 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/app/ui/toc/symbology-stack.directive.js, line 132 at r3](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-KwAJ9rK4jZxiN5sW8qL:-KwB-3J6IREGagfVlAPR:bb6un1j) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/5c6e3677bded78f9a61f0877e7b01dc28f44f27f/src/app/ui/toc/symbology-stack.directive.js#L132)):*

Can you add this event to the events list to avoid hardcoding random strings?
</blockquote></details>

Done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwB--KDO4HxVgFEheIS:b-at7a4f)*

AleksueiR:
When you collapse the stack, the checkbox is visible on top:


When expanding, the checkboxes are rendered on top of symbols - they should be underneath them:


Also, do we actually want to show a symbology checkbox if there only one item, like in here:


It just mimics the visiblity toggle at this point.

---

Reviewed 2 of 3 files at r4.
Review status: 9 of 10 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwBFpdrv8g6JWUDk2-D:b1bgdn4)*

milesap:
Removed toggles for one symbol layers and fixed that pesky lingering checkbox

---

Review status: 9 of 10 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwBMjLc5wtbsTy-VVjo:b-oxacsx)*

AleksueiR:
The pesky things are still there, but they are getting hidden after the transition ends. Can you move the checkboxes underneath the images, so checkboxes won't overlay the images at the start and end of the transition?



The checkboxes are still a tiny bit off alignment, but they are getting there:


---

Review status: 9 of 10 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwCGkvz7gKIlD7Yluth:b-xfoye8)*

james-rae:
Reviewed 1 of 2 files at r3, 2 of 3 files at r4, 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418)*

milesap:
Alignment should be pixel perfect now. I can't make the checkbox go away before the transition ends because:
- z-index changes cause more problems
- symbol placement after the checkbox breaks layout
- **TimelineLite does not have a onReverse event??** so I can't know when a reverse animation has started. https://greensock.com/docs/TimelineLite

---

Review status: 9 of 10 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwGNa1ETFyfDRapz3Vy:b-1l32h3)*

james-rae:
Legend text seems to be chopping early (this is sea parks layer)


---







*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwGlDwhSk9QDLohfceD:b-wnqrnp)*

james-rae:
A bit of mis-match between the legend filters and the APPLY button in the datagrid.

Case 1: After adding a grid filter and clicking apply on the datagrid, then changing the legend filter, the applied filter is lost on the map.  This might be an OK behaviour, as the Apply button is sort of a one-way street. But we could also look at preserving any existing grid filter when applying legend filters. Legend guts would need a way to ask the grid if it has an active filter, and get the query string from it  (connect them with an OR i think?)

Case 2: Do some legend filters. Then do a grid filter and click apply.  The map loses the legend filters. This one is probably bad because the map doesn’t match the checkboxes in the legend. So a bit of the same, when applying the grid filter, need to tack on any legend filters that are existing.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwGo4i7xymF5IrPoMpJ:b-7570fz)*

james-rae:
My bad, the legend pic above was not sea parks, it was the [incident service](http://geoappext.nrcan.gc.ca/arcgis/rest/services/NEB/Incident/MapServer/0)

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwGtys2qr1RmEeoBK_-:bf0cj0l)*

james-rae:
In the legend, when a grid filter is applied to the map, a small funnel icon appears on the legend entry with hovertext `Some features are hidden by filters`. Should we also apply this icon when the symbol filter is on?  Or conversely, change the hovertext to `Some features are hidden by table filters` (I like this one, since the checkboxes on the legend already give a visual queue to the symbol filter)

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwH3KxRsvqZBHcHvw-x:b8d8edo)*

AleksueiR:
Why do you need the `onReverse` event? Just call whatever you want after you call `.reverse();` to reverse the animation.

---

Review status: 9 of 10 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwZxoKzEKt3gAccol5g:badvg8g)*

milesap:
To be fair they have `onStart` so omitting `onReverse` is a head scratcher 

---

Review status: 8 of 10 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-Kw_1K7F-Ufw34pzIxgi:byw0mq6)*

AleksueiR:
Ok, if you want to show the checkboxes after the transition, can you add a small fade in/out animation to smooth things out a bit?


There are two more small problems: the "Zone" label is being chopped off and the close "X" button for the symbology stack is pushed to the left compared to the current implementation:

| new                                  | old                                  |
| ------------------------------------ | ------------------------------------ |
|  |  |

---

Review status: 8 of 10 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-Kw_2e-4nn0psFL0dYHO:be4f4al)*

milesap:
Think we're good now :crossed_fingers: 

---

Review status: 7 of 10 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-Kw_JRErwOnOIUeKr_Wg:b-o3zw78)*

AleksueiR:
Almost :D The close button is too much to the right now (maybe three or four pixels) :D we are getting there.

##### new:


##### old:


---

Reviewed 3 of 3 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwaBl_XWZE0Uy-hEIRy:b-t4rx7n)*

milesap:
I'm so close, I think I can see the light at the end of the tunnel :sun_with_face: 

---

Review status: 9 of 10 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwaEA3Xyt1Xhhp7DI0-:bvtk0cg)*

AleksueiR:
Nice :lgtm: 

---

Reviewed 1 of 1 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwaFJxOPKpsy_n0vCH5:bzhmkty)*

JahidAhmed:
Small thing I noticed, and steps to reproduce (on Sample 46 Dynamic Layer with only one child with filter -> Power Plants):
Deselect everything other than `Biomass` and `Coal` and filter the table to for example 3.
Now deselect `Coal`, all the remaining `Biomass` features appear on the map that weren't there before, but not in the table since they are filtered.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwaHekpOck3JKR_FYKv:b94tl2q)*

JahidAhmed:
Clearing filters will also display all the features, even if they aren't selected. 

And also, do we want to display the filter flag in cases where we aren't showing all the symbology?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwaIMJUQ7UI0XJJX6Wl:b-qncyvq)*

milesap:
Good points -  having symbol toggle and table filtering to work together would take (much?) more work but we can improve on this if the need arises (maybe for CESI). 

I'm ok with no filter flag but will add it if people complain :sun_with_face: 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwaTQhJ2HHe9191056T:b-d654pn)*

AleksueiR:
Some comments up, @james-rae suggested to change the filter flag's message to `Some features are hidden by table filters` since the toggles on the symbology stack give indication that features are filtered.

Argh, I think the focus manager needs to be made aware of all these extra toggles. You kinda can tab through them but only if you start tabbing backwards after you open the stack...

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwjKDQGSya-2pjmTQ9R:b64fb8x)*

milesap:
Changed the filter flag text. 

Since symbology is rendered backwards in the DOM (close button is after symbols, and symbols are in reverse order) FM will now tab in reverse order starting at the bottom. On the plus side it seems like a fancy feature we spent a lot of time making. 

---

Review status: 7 of 10 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwjTNlm56TkKGQpaR6B:b-8c0p4x)*

AleksueiR:
When tabbing, the focus gets stack on the symbology stack and doesn't move forward. (index-samples, config 1)


---

Review status: 7 of 10 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwoRRSCaAjIpFjSWv5A:buq41y2)*

AleksueiR:
re: backward rendered symbology
Here's the double-reversed version: https://github.com/AleksueiR/fgpv-vpgf/commit/03937639d8433c331007cc130a281b5cc311e06e

and the demo: http://fgpv.cloudapp.net/demo/users/AleksueiR/2368-reverse-symbology-stack-dom-order/prod/samples/index-samples.html

---

Review status: 7 of 10 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwoT-1daEzuLLo0_Kfd:b5abaph)*

james-rae:
Everything seems broken to me now.  I also no longer see `package.json` in the file list...did we lose our pointer to the matching geoApi?

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwpAVQdgnC6yets81FH:b8bvpii)*

milesap:
Fixed mysterious package.json disappearance - should be working again. 

Implemented your symbology reorder @aleksueir 

Focus manager is happy once more

---

Review status: 5 of 12 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-KwpgLAehKb41LZzQKsx:b6q8bnq)*

james-rae:
K now I need to make a change to geoApi. Turns out the where clause in a query is data-type agnostic, but in map image export it's data-type sensitive.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-Kwq5T23rV_fo8OArHPV:b98q1ry)*

james-rae:
Reviewed 2 of 3 files at r6, 1 of 1 files at r7, 1 of 2 files at r8, 2 of 2 files at r9, 5 of 5 files at r10.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418)*

milesap:
Looks like we're good to go, final geoApi implemented  and @james-rae opened future issues to hash out:

fgpv-vpgf/fgpv-vpgf#2438
fgpv-vpgf/fgpv-vpgf#2439

---

Review status: 8 of 12 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-Kx7wCc0BLmEkESQlPK8:b5tx9fh)*

AleksueiR:
:lgtm:  :+1: 

good to merge?

---

Reviewed 1 of 2 files at r9, 3 of 5 files at r10, 3 of 3 files at r11, 1 of 1 files at r12.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-Kx8CGWpRreKCX3dxcbw:bxz2rem)*

milesap:
Good to go

milesap:
Good to go

AleksueiR:
I don't think this is a good idea to point to the geoapi's develop branch. This can be prone to breaking if someone will merge geoapi changes that are not yet supported by the client.

---

Reviewed 1 of 1 files at r13.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418#-:-Kx8FkI7r6H2qOjCE8o0:b8zqcnb)*

AleksueiR:
Reviewed 1 of 1 files at r14.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2418)*

Issue #1985, opened 2017-06-16
Be able to easily enable/disable identify


jolevesq:
For example, with the plugin coord info we would like to disable the identify task when user click on map to get coordinate info the setting identify back when user close the information panel.

mweech:
Thoughts:
- have a config based toggle for default behavior..
- also provide a run-time internal toggle a plug-in could use to suppress built-in viewer Identify
- how do we deal with poor plug-ins?  could suppress and then screw up and never re-enable.. perhaps some kind of notice to the user while it's disabled..? "plug-in has temporarily disabled Map Identify.  Click here to re-enable"

AleksueiR:
A runtime `identify` toggle will be implemented as part of the following issue: https://github.com/fgpv-vpgf/fgpv-vpgf/issues/2595.

AleksueiR:
@james-rae Can we close this one?

PR #2597, opened 2018-03-15
feat(identify): improvements to identify flow

### LINKED ISSUE #2595: Identify Highlight Improvements
## CIP Requirement
Highlight features on the map without opening the details panel.

## Default behaviour
The identify logic is closely tied with the details panel:
1. clicking on the map triggers identify on all layers
2. IdentifyResults from each layers (Dynamic layers have one for each sublayer) are grouped
3. and, when resolved, results are passed to the details panel
4. the panel breaks the results by originating sublayer and renders on the screen (mini sublayer selector, and the results for the selected sublayer)
5. a single subset of results is selected (either first one with data, or cached user choice)
6. only the results __visible__ in the details panel (from a single sublayer) are highlighted on the map

The actual highlighting happens at the very end of the `identify` flow and the highlighted items depend on the order in which the results are returned or cached user choice. 

## Improvements

Define additional identify flows to satisfy the requirement and not do a massing refactor of the identify/details modules. Four identify modes:

### `Details` aka `default behaviour`
Runs identify query, highlights (only ones from a visible sublayer in the details panel) the results and displays them in the details panel.

### `Highlight` aka `cip requirement`
Runs identify query, and highlights all results on the map.

### `Silent`
Runs identify query only.

### `None` aka https://github.com/fgpv-vpgf/fgpv-vpgf/issues/1985
No identify. 

Example:
```ts
RZ.mapInstances[0].identifyMode = 'details';
RZ.mapInstances[0].identifyMode = 'highlight';
RZ.mapInstances[0].identifyMode = 'silent';
RZ.mapInstances[0].identifyMode = 'none';
```

## Also

Identify results will be piped through the `identify` API endpoint Observable for the host page to consume. The host page, can modify/remove/add identify results subscribing to this stream before the results will be processed by RAMP. If the items are removed (for example), they will not be highlighted or shown in the details panel.

How can this be useful? 

- If the host page wants to run the identify query and then do something else with the results without affecting RAMP, they can set the `silent` mode.

- CIP needs to highlight a single grid cell (CMIP5 dataset) on user click, but the default behaviour can possible highlight two or three or four cells depending how close to the cell boundary the user is clicking. To avoid this, all except one result can be removed by the host page using the `identify` observable which will only highlight one feature. The `highlight` mode also needs to be set.

Example:
```ts
RZ.mapInstances[0].identify.subscribe(identifyResults =>
	identifyResults.forEach(result => {
		if (result.data.length > 0) {
			result.data.splice(1);
		}
	})
);
```

AleksueiR:
## Description
Specifies four different identify modes: details, highlight, silent and none. See #2595 for more details.
Setting identify mode to `none` can disable the identify functionality at runtime as described in #1985.

### How to test
1. load `index-samples` page
2. type `RZ.mapInstances[0].identifyMode` in the console // identify mode defaults to 'details'
3. click on things, see them get highlighted and displayed in the details panel
4. type the following in the console (this will remove all but one result from each sublayer)
```ts
RZ.mapInstances[0].identify.subscribe(identifyResults =>
	identifyResults.forEach(result => {
		if (result.data.length > 0) {
		        result.data.splice(1);
		}
	})
);
```
5. click on things, see only one feature get highlighted and displayed in the details panel
6. type `RZ.mapInstances[0].identifyMode = 'highlight';` in the console
7. click on things - they are only highlighted now (still only one feature because the filter is still working)
8. type `RZ.mapInstances[0].identifyMode = 'silent';` in the console
8. type the following in the console
```ts
RZ.mapInstances[0].identify.subscribe(identifyResults => console.log(identifyResults));
```
10. click on things - see nothing happen on the map, but the results are printed in the console still
11. type `RZ.mapInstances[0].identifyMode = 'none';` in the console
12. click on this - notice nothing is happening
13. eat a muffin

Closes #2595 
Relates #1985 

## Testing

:eyes: 

## Documentation

Inline comments.

## Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/2597)

### DIFF
--- a/api/src/layers.ts
+++ b/api/src/layers.ts
@@ -941,11 +941,13 @@ export interface IdentifyResult {
 
 export interface IdentifyRequest {
     sessionId: number;
+    event: MouseEvent
     layer: ConfigLayer;
     features: Promise<IdentifyResult[]>;
 }
 
 export interface IdentifySession {
     sessionId: number;
+    event: MouseEvent
     requests: IdentifyRequest[];
 }
--- a/api/src/layers.ts
+++ b/api/src/layers.ts
@@ -695,6 +695,7 @@ export class LayerGroup {
     _attributesRemoved: Subject<LayerAndAttribs>;
 
     _click: Subject<BaseLayer>;
+    
     /** @ignore */
     _identify: Subject<any>;
 
@@ -769,7 +770,7 @@ export class LayerGroup {
     /**
      * Emits when a new identify session as soon as the identify is triggered.
      * The emitted event contains an array of IdentifyRequest objects, one for each layer in this layer group.
-     * 
+     *
      * @readonly
      * @type {Observable<IdentifySession>}
      * @event identify
@@ -932,22 +933,27 @@ interface LayerAndChangedAttribs extends ChangedAttribs {
     layer: BaseLayer;
 }
 
+export interface DataItem {
+    name: string;
+    value: string | number;
+}
+
 export interface IdentifyResult {
-    data: RV.LAYER.DataItem[],
-    name: string,
-    oid: string,
-    symbology: { svgcode: string }[]
+    data: DataItem[];
+    name: string;
+    oid: string;
+    symbology: { svgcode: string }[];
 }
 
 export interface IdentifyRequest {
     sessionId: number;
-    event: MouseEvent
+    event: MouseEvent;
     layer: ConfigLayer;
     features: Promise<IdentifyResult[]>;
 }
 
 export interface IdentifySession {
     sessionId: number;
-    event: MouseEvent
+    event: MouseEvent;
     requests: IdentifyRequest[];
 }
--- a/api/src/layers.ts
+++ b/api/src/layers.ts
@@ -20,6 +20,7 @@ import { BehaviorSubject } from 'rxjs/BehaviorSubject';
 import { DynamicLayerEntryNode, InitialLayerSettings } from 'api/schema';
 import { BaseGeometry, Point } from 'api/geometry';
 import Map from 'api/map';
+import { RV } from './index';
 
 const layerTypes = {
     ESRI_GRAPHICS: 'esriGraphics',
@@ -28,7 +29,7 @@ const layerTypes = {
     ESRI_IMAGE: 'esriImage',
     ESRI_TILE: 'esriTile',
     OGC_WMS: 'ogcWms'
-}
+};
 
 /**
  * All layer types must derive from this class. Not intented to be instantiated on its own.
@@ -83,9 +84,9 @@ export class BaseLayer {
         this._opacityChanged = new Subject();
         this._visibilityChanged = new Subject();
 
-        this._nameChanged.subscribe(name => this._name = name || '');
-        this._opacityChanged.subscribe(opacity => this._opacity = opacity);
-        this._visibilityChanged.subscribe(visibility => this._visibility = visibility);
+        this._nameChanged.subscribe(name => (this._name = name || ''));
+        this._opacityChanged.subscribe(opacity => (this._opacity = opacity));
+        this._visibilityChanged.subscribe(visibility => (this._visibility = visibility));
 
         this._attributeArray = [];
         this._attributesAdded = new BehaviorSubject(this._attributeArray);
@@ -144,7 +145,7 @@ export class BaseLayer {
     }
 
     /** Forces an attribute download. Function implementation in subclasses. */
-    fetchAttributes(): void { }
+    fetchAttributes(): void {}
 
     /** Sets the attribute object to value provided using the attributeKey. */
     setAttributes(attributeKey: number, value: Object): void;
@@ -155,8 +156,9 @@ export class BaseLayer {
     /** Sets the entire attribute or sets an individual field inside attribute using attributeKey provided. */
     setAttributes(attributeKey: number, valueOrFieldName: Object | string, value?: string | number): void {
         if (typeof valueOrFieldName === 'string') {
-            let attribValue: Object | undefined = this._attributeArray.find(attrib =>
-                (<any>attrib)[this._primaryAttributeKey] === attributeKey);
+            let attribValue: Object | undefined = this._attributeArray.find(
+                attrib => (<any>attrib)[this._primaryAttributeKey] === attributeKey
+            );
 
             if (typeof attribValue !== 'undefined') {
                 const oldValue: Object = Object.assign({}, attribValue);
@@ -165,8 +167,9 @@ export class BaseLayer {
                 this._attributesChanged.next({ attributesBeforeChange: oldValue, attributesAfterChange: attribValue });
             }
         } else {
-            let index: number = this._attributeArray.findIndex(attrib =>
-                (<any>attrib)[this._primaryAttributeKey] === attributeKey);
+            let index: number = this._attributeArray.findIndex(
+                attrib => (<any>attrib)[this._primaryAttributeKey] === attributeKey
+            );
 
             if (index !== -1) {
                 const oldValue: Object = Object.assign({}, this._attributeArray[index]);
@@ -175,19 +178,28 @@ export class BaseLayer {
                     (<any>this._attributeArray[index])[key] = (<any>valueOrFieldName)[key];
                 });
 
-                this._attributesChanged.next({ attributesBeforeChange: oldValue, attributesAfterChange: this._attributeArray[index] });
+                this._attributesChanged.next({
+                    attributesBeforeChange: oldValue,
+                    attributesAfterChange: this._attributeArray[index]
+                });
             }
         }
     }
 
     /** Returns the layer ID. */
-    get id(): string { return this._id; }
+    get id(): string {
+        return this._id;
+    }
 
     /** Returns the layer index. */
-    get layerIndex(): number | undefined { return this._layerIndex; }
+    get layerIndex(): number | undefined {
+        return this._layerIndex;
+    }
 
     /** Returns the name of the layer. */
-    get name(): string { return this._name; }
+    get name(): string {
+        return this._name;
+    }
 
     /** Sets the name of the layer. This updates the name throughout the viewer.
      *
@@ -224,7 +236,9 @@ export class BaseLayer {
     }
 
     /** Returns the opacity of the layer on the map from 0 (hidden) to 100 (fully visible). */
-    get opacity(): number { return this._opacity; }
+    get opacity(): number {
+        return this._opacity;
+    }
 
     /** Sets the opacity value for the layer. */
     set opacity(opacity: number) {
@@ -235,7 +249,9 @@ export class BaseLayer {
 
         if (this._viewerLayer.config) {
             if (this._layerIndex !== undefined) {
-                const childNode = this._viewerLayer.config.layerEntries.find((l: DynamicLayerEntryNode) => l.index === this._layerIndex);
+                const childNode = this._viewerLayer.config.layerEntries.find(
+                    (l: DynamicLayerEntryNode) => l.index === this._layerIndex
+                );
                 childNode.state.opacity = opacity;
             } else {
                 this._viewerLayer.config.state.opacity = opacity;
@@ -256,7 +272,9 @@ export class BaseLayer {
     }
 
     /** Returns true if the layer is currently visible, false otherwise. */
-    get visibility(): boolean { return this._visibility; }
+    get visibility(): boolean {
+        return this._visibility;
+    }
 
     /** Sets the visibility to visible/invisible. */
     set visibility(visibility: boolean) {
@@ -267,7 +285,9 @@ export class BaseLayer {
 
         if (this._viewerLayer.config) {
             if (this._layerIndex !== undefined) {
-                const childNode = this._viewerLayer.config.layerEntries.find((l: DynamicLayerEntryNode) => l.index === this._layerIndex);
+                const childNode = this._viewerLayer.config.layerEntries.find(
+                    (l: DynamicLayerEntryNode) => l.index === this._layerIndex
+                );
                 childNode.state.visibility = visibility;
             } else {
                 this._viewerLayer.config.state.visibility = visibility;
@@ -292,8 +312,9 @@ export class BaseLayer {
         if (typeof attributeKey !== 'undefined') {
             let allAttribs: Array<Object> = this.getAttributes();
 
-            let index: number = this._attributeArray.findIndex(attrib =>
-                (<any>attrib)[this._primaryAttributeKey] === attributeKey);
+            let index: number = this._attributeArray.findIndex(
+                attrib => (<any>attrib)[this._primaryAttributeKey] === attributeKey
+            );
 
             if (index !== -1) {
                 const oldValue: Object = Object.assign({}, this._attributeArray[index]);
@@ -304,7 +325,7 @@ export class BaseLayer {
 
                 allAttribs.splice(index, 1);
 
-                this._attributesRemoved.next([ oldValue ]);
+                this._attributesRemoved.next([oldValue]);
             }
         } else {
             const copyAttribs: Array<Object> = this._attributeArray.map(a => Object.assign({}, a));
@@ -386,41 +407,49 @@ export class ConfigLayer extends BaseLayer {
         const attribs = this._layerProxy.attribs;
 
         if (attribs) {
-            attribs.then((attrib: AttribObject) => {
-                // the attributes were previously downloaded, do not reupdate the array and do not trigger `attributes_added`
-                if (this._attributeArray.length > 0) {
+            attribs
+                .then((attrib: AttribObject) => {
+                    // the attributes were previously downloaded, do not reupdate the array and do not trigger `attributes_added`
+                    if (this._attributeArray.length > 0) {
+                        return;
+                    }
+
+                    // attributes not previously downloaded, after forcing the download, populates the array and triggers event
+                    Object.keys(attrib.oidIndex).forEach(id => {
+                        const index: number = (<any>attrib.oidIndex)[id];
+                        const attribs = attrib.features[index].attributes;
+
+                        this._attributeArray.push(attribs);
+                    });
+
+                    this._attributesAdded.next(this._attributeArray);
+                })
+                .catch((e: string) => {
+                    console.error(e);
+                    this._attributeArray = [];
+                    this._attributesAdded.next(this._attributeArray); // errored out, do we want to broadcast a different event  ?
                     return;
-                }
-
-                // attributes not previously downloaded, after forcing the download, populates the array and triggers event
-                Object.keys(attrib.oidIndex).forEach(id => {
-                    const index: number = (<any>attrib.oidIndex)[id];
-                    const attribs = attrib.features[index].attributes;
-
-                    this._attributeArray.push(attribs);
                 });
-
-                this._attributesAdded.next(this._attributeArray);
-            }).catch((e: string) => {
-                console.error(e);
-                this._attributeArray = [];
-                this._attributesAdded.next(this._attributeArray);   // errored out, do we want to broadcast a different event  ?
-                return;
-            });
         } else {
             this._attributeArray = [];
-            this._attributesAdded.next(this._attributeArray);   // no attribs, do we want to broadcast a different event  ?
+            this._attributesAdded.next(this._attributeArray); // no attribs, do we want to broadcast a different event  ?
         }
     }
 
     /** Returns the catalogue URL. */
-    get catalogueUrl(): string { return this._catalogueUrl; }
+    get catalogueUrl(): string {
+        return this._catalogueUrl;
+    }
 
     /** Returns the underlying layer type such as esriFeature, esriDynamic, and ogcWms. */
-    get type(): string { return this._layerType; }
+    get type(): string {
+        return this._layerType;
+    }
 
     /** Returns the name of the key being used for the attributes OID field. */
-    get attributeKey(): string { return this._primaryAttributeKey; }
+    get attributeKey(): string {
+        return this._primaryAttributeKey;
+    }
 
     /** Pans to the layers bounding box. */
     panToBoundary(): void {
@@ -509,7 +538,9 @@ export class SimpleLayer extends BaseLayer {
     }
 
     /** Returns the name of the layer. */
-    get name(): string { return this._name; }
+    get name(): string {
+        return this._name;
+    }
 
     /** Sets the name of the layer. Will also update the id to reflect this change, since the id is equivalent to the name. */
     set name(name: string) {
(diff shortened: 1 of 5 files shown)


james-rae:
Random thought.  In the `identify` observable on the API, should the `.requester` property be mapped to an API class (e.g. `ConfigLayer`) instead of the `LayerInterface`?  I'm assuming at this point `LayerInterface` is considered an internal class?
We can do this in another change if CIP is blocked by not having identify api enhancements.

---

Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[src/app/geo/identify.service.js, line 97 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597#-L7ekvG6rOX18sJ414M6:-L7ekvG6rOX18sJ414M7:bp9coq5) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/711a17bc5bc62f91783d4156e08703ca0b74348a/src/app/geo/identify.service.js#L97)):*
> ```JavaScript
>             // they will not be highlighted or shown in the details panel
>             mApi._identifySubject.next(allIdentifyResults);
>             return true;
> ```

does this mean our API will not tell about identify results until every layer has reported back or timed out?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597#-:-L7ejwuyEO2wP6tME3G_:b-72gl3a)*

AleksueiR:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[src/app/geo/identify.service.js, line 97 at r1](https://beta.reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/2597#-L7ekvG6rOX18sJ414M6:-L7ewDmfEDihiI8tBv67:b-rlxyhj) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/711a17bc5bc62f91783d4156e08703ca0b74348a/src/app/geo/identify.service.js#L97)):*

does this mean our API will not tell about identify results until every layer has reported back or timed out?
</blockquote></details>

Correct. We can return results from a single layer as soon as any if available, but then it becomes harder to say which results belong to which identify query - you would need something like a `start/stop identify` events to mark the beginning and completing of the identify request. This is definitely doable, but is there a use case for this? 

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/2597)*

AleksueiR:
That's a good point. `LayerInterface` represents a proxy layer object created by geoapi. Is it going to be complicated to map them to API classes?

---

Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/2597#-:-L7ewwPP4SbVWAiF-o_P:b-9tq9ih)*

james-rae:
@jahidahmed can you give your expert opinion?

---

Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[src/app/geo/identify.service.js, line 97 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597#-L7ekvG6rOX18sJ414M6:-L7f3lQJ2_QlwphDaXHV:bx6hbck) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/711a17bc5bc62f91783d4156e08703ca0b74348a/src/app/geo/identify.service.js#L97)):*

Correct. We can return results from a single layer as soon as any if available, but then it becomes harder to say which results belong to which identify query - you would need something like a `start/stop identify` events to mark the beginning and completing of the identify request. This is definitely doable, but is there a use case for this? 
</blockquote></details>

Both approaches have advantages. 

Wait for everything approach is easier to manage in code. You can assume an atomic result and exploit that.

Results as they occur approach offers better protection against hanging servers, and gets results to the user faster (think of how nice the RAMP identify pane is, with the spinners resolving on each layer). But it requires much more effort to manage.  Multiple promises in multiple places; you would need an identify session id or start/stop as you suggested; highlight all needs to worry about things getting cleared halfway through the identify.

I can see use-cases for both -- it mainly depends on the data.  Wait for everything works good in CIP, as its generally one layer on a reliable server.  FGP shopping cart map is much more tailored to results as they occur.

This reply doesn't give a path forward :trophy:  , but maybe we can haggle in the morning.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597#-:-L7f52j3uINdSpADi9R1:b-10e9u0)*

AleksueiR:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[src/app/geo/identify.service.js, line 97 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597#-L7ekvG6rOX18sJ414M6:-L7iVbisLSRXIXwcyuA-:b-ujvhba) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/711a17bc5bc62f91783d4156e08703ca0b74348a/src/app/geo/identify.service.js#L97)):*

Both approaches have advantages. 

Wait for everything approach is easier to manage in code. You can assume an atomic result and exploit that.

Results as they occur approach offers better protection against hanging servers, and gets results to the user faster (think of how nice the RAMP identify pane is, with the spinners resolving on each layer). But it requires much more effort to manage.  Multiple promises in multiple places; you would need an identify session id or start/stop as you suggested; highlight all needs to worry about things getting cleared halfway through the identify.

I can see use-cases for both -- it mainly depends on the data.  Wait for everything works good in CIP, as its generally one layer on a reliable server.  FGP shopping cart map is much more tailored to results as they occur.

This reply doesn't give a path forward :trophy:  , but maybe we can haggle in the morning.
</blockquote></details>

All fair points :+1: 

There is a third alternative - we push the full identify package as it is into the pipe as soon as all the promises are assembled (not resolved), and then the host page can decide whether to wait on all promises to clear or process them as they resolve one by one. The host page wouldn't need much more code than what is needed right now (no need for `start/end` events or to track session ids) and the implementation on the RAMP side is not lock into a single scenario.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597)*

spencerwahl:
Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597)*

JahidAhmed:
I think in order to map the `LayerInterface` to the API Layers, it would be something similar to what is happening in this block [here](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/develop/src/app/geo/legend-block.class.js#L281) for dynamic layers; we have to compare the entire proxy objects in order to find the matching api layer (explained somewhat in comment above code) since our `LayerInterface` doesn't have any ID and index fields. Not sure if that's the best option, but I think we do want to look at adding the fields mentioned to the proxy object

---

Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597#-:-L7if0ACSOwT8DwkT5UY:b-cowf5g)*

AleksueiR:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[src/app/geo/identify.service.js, line 97 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597#-L7ekvG6rOX18sJ414M6:-L7ipukoflLeO3kY7lvE:ba7d5bi) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/711a17bc5bc62f91783d4156e08703ca0b74348a/src/app/geo/identify.service.js#L97)):*

All fair points :+1: 

There is a third alternative - we push the full identify package as it is into the pipe as soon as all the promises are assembled (not resolved), and then the host page can decide whether to wait on all promises to clear or process them as they resolve one by one. The host page wouldn't need much more code than what is needed right now (no need for `start/end` events or to track session ids) and the implementation on the RAMP side is not lock into a single scenario.
</blockquote></details>

Upon some further pondering, I would change the structure of the `identifyPackage` going into the observable pipe to make it simpler for the host page to consume. 

```ts
interface IdentifyPackage {
    requester: {
        mapPoint: MapPoint
    },
    requests: {
        requester: LayerInterface,
        results: Promise<{ name: string, data: any[], oid: string, symbology: any[] }[]>
    }[]
}
```

Here, `requests` is an array of objects with two properties: a `requester` object containing the proxy object and an actual `results` promise returning the identify results.

This is a bit different from what RAMP is uses internally. Internally, we have a list of `identifyInstances`, one per layer. Each instance has a single `identifyPromise`, and a list of `identifyResults` (multiple in cases of dynamic layer - one per sublayer) which contain identify data arrays. 

For the purposes of simplification, we map this combined array of `identifyResults` to the array of `requests` described above. For dynamic layer, the same single `identifyPromise` will be use in with several `identifyResults` in corresponding `request` objects.

This conversion can be done with something like this:

```ts
const requests = identifyInstances.reduce((map, { identifyPromise, identifyResults }) => {
    const instanceRequests = identifyResults.map(r => ({
        requester: r.requester.proxy,
        results: identifyPromise.then(() => r.data)
    }));
    return map.concat(instanceRequests);
}, []);

mApi._identifySubject.next({
    requests,
    requester
});
```

Then, the host page can do the following:

```ts
const t0 = performance.now();

// watching results as they come in
RZ.mapInstances[0].identify.subscribe(event => {
    console.log('%c Parsing identify results as they resolve', 'background: #222; color: #bada55');
    
    event.requests.forEach((r, index) => r.results.then(data => console.log(index, r.requester.name, data, `resolved in ${(performance.now() - t0).toFixed(2)} milliseconds`)));
});

// waiting for all results to clear
RZ.mapInstances[0].identify.subscribe(event => {
    
    Promise.all(event.requests.map(r => r.results)).then(allResults => {
        console.log('%c Waiting for all identify results to resolve first', 'background: #222; color: #bada55');
        const t1 = performance.now() - t0;
        
        allResults.forEach((data, index) => console.log(index, event.requests[index].requester.name, data, `resolved in ${t1.toFixed(2)} milliseconds`));
  
        // we can still easily remove all but one results
        // allResults.forEach(data => data.splice(1));
    });
});
```

The above code produces this cheerful output:


The timings of result resolutions indicate that they are indeed resolved in the order desired.

Thoughts?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597)*

AleksueiR:
@jahidahmed Thank you for the explanation. This does seem like a sizable chunk of work though, and should be spun off as a separate issue. I don't believe CIP will need to use the `requester`'s proxy object much if at all, so any future changes to what is returned as the layer's proxy won't affect CIP code much.

---

Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597#-:-L7irQfhZms0VU_B58FS:b-ehn4nl)*

james-rae:
Thanks Jahid.  Agree, something to change later

---

Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[src/app/geo/identify.service.js, line 97 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597#-L7ekvG6rOX18sJ414M6:-L7j5FoFXmLI43bpVakp:b-cu140c) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/711a17bc5bc62f91783d4156e08703ca0b74348a/src/app/geo/identify.service.js#L97)):*

Upon some further pondering, I would change the structure of the `identifyPackage` going into the observable pipe to make it simpler for the host page to consume. 

```ts
interface IdentifyPackage {
    requester: {
        mapPoint: MapPoint
    },
    requests: {
        requester: LayerInterface,
        results: Promise<{ name: string, data: any[], oid: string, symbology: any[] }[]>
    }[]
}
```

Here, `requests` is an array of objects with two properties: a `requester` object containing the proxy object and an actual `results` promise returning the identify results.

This is a bit different from what RAMP is uses internally. Internally, we have a list of `identifyInstances`, one per layer. Each instance has a single `identifyPromise`, and a list of `identifyResults` (multiple in cases of dynamic layer - one per sublayer) which contain identify data arrays. 

For the purposes of simplification, we map this combined array of `identifyResults` to the array of `requests` described above. For dynamic layer, the same single `identifyPromise` will be use in with several `identifyResults` in corresponding `request` objects.

This conversion can be done with something like this:

```ts
const requests = identifyInstances.reduce((map, { identifyPromise, identifyResults }) => {
    const instanceRequests = identifyResults.map(r => ({
        requester: r.requester.proxy,
        results: identifyPromise.then(() => r.data)
    }));
    return map.concat(instanceRequests);
}, []);

mApi._identifySubject.next({
    requests,
    requester
});
```

Then, the host page can do the following:

```ts
const t0 = performance.now();

// watching results as they come in
RZ.mapInstances[0].identify.subscribe(event => {
    console.log('%c Parsing identify results as they resolve', 'background: #222; color: #bada55');
    
    event.requests.forEach((r, index) => r.results.then(data => console.log(index, r.requester.name, data, `resolved in ${(performance.now() - t0).toFixed(2)} milliseconds`)));
});

// waiting for all results to clear
RZ.mapInstances[0].identify.subscribe(event => {
    
    Promise.all(event.requests.map(r => r.results)).then(allResults => {
        console.log('%c Waiting for all identify results to resolve first', 'background: #222; color: #bada55');
        const t1 = performance.now() - t0;
        
        allResults.forEach((data, index) => console.log(index, event.requests[index].requester.name, data, `resolved in ${t1.toFixed(2)} milliseconds`));
  
        // we can still easily remove all but one results
        // allResults.forEach(data => data.splice(1));
    });
});
```

The above code produces this cheerful output:


The timings of result resolutions indicate that they are indeed resolved in the order desired.

Thoughts?

</blockquote></details>

I really like this approach.  Maximum flexibility with minimal complexity on the API. Outstanding.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597#-:-L7j5NQKKsmKmdega1u6:bkodmvj)*

milesap:
Everything works well, though we prob. need to discuss what the `identify` observable is emitting; Here's my 3 cents:

Cent #1: The`identify` property on a map instance should be moved to `ConfigLayer`, and aliased on `LayerGroup`. `LayerGroup` would not emit for `SimpleLayer` since we don't support identify on `SimpleLayer`'s. `ConfigLayer` would receive an identifyResults attribute which is a subset of it's full data set (its location results).

> This allows you to subscribe to all `ConfigLayer`'s or only ones you're interested in directly on the layer itself, or via the `LayerGroup` (all at once).

Cent #2: The `identify` property on `LayerGroup` should emit a `ConfigLayer` instance as soon as that layer has data. So for example

```js
RZ.mapInstances[0].layers.identify.subscribe(confLayer => {
    console.log(confLayer.identifyResults);
});

```

> When a new identify is triggered/clicked by user, `identify` emits a falsy value (`undefined`, or `null` or `false`). This clears `identifyResults` can also acts as a notification that subsequent emits are from a new identify. 

Cent #3:  Likewise the `identifyMode` property should be moved to `ConfigLayer`, and aliased on `LayerGroup`.

Let me know if anything I said doesn't make much sense or you have alternate ideas. If this is too much work you can open a new issue for these changes and we'll likely schedule them in for a next sprint completion time.

---

Reviewed 2 of 4 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/2597#-:-L7iwGc607gwKMFtGcLp:b-117hkx)*

james-rae:
Identify is tricky. The results come from layers, but the action is applied to the map.

For cent 1, this seems ok to be as it gives options to easily target all results or an individual layers results. I would like to see Alex's suggestion in the comments below be preserved in this.

For cent 2, this kind of goes against the suggestion below in comments. I'd prefer the `LayerGroup` still emit the block of promises, while the `ConfigLayer`s can emit things as they arrive.  Alternatively, now that we have two endpoints, we can entertain reverting back to the `LayerGroup` not emitting anything until everything is done, then emit all results instead of promises.  We can haggle more on this.

For cent 3, this gets a bit tricky, as two of the modes affect ALL results (Detail and Highlight), so it makes sense to be on the map.  But it's also useful to be able to maybe exclude some layers from results. But I think there is already a `queryable` property that does this, so perhaps best to use that intead.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597#-:-L7jUNfqglJcdgKihmL1:b21ab5k)*

milesap:
Looks like another meeting is in order, I'll send one out shortly.

---

Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597#-:-L7k8r-N6hvgVIG0f8IJ:b2tyoga)*

milesap:
Per today's impromptu meeting this is good to go, and I'll open a follow up PR soon for the changes.

:lgtm: 

---

Reviewed 2 of 4 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597#-:-L7z3actbVUVCetPcddG:bxwfe91)*

AleksueiR:
Implement the `layerGroup` identify flow part from https://github.com/fgpv-vpgf/fgpv-vpgf/issues/2601. 

---

Review status: 2 of 5 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597#-:-L82WJXiO6hi-46nhw4L:b-jqmsmy)*

AleksueiR:
Review status: 2 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/app/geo/identify.service.js, line 97 at r1](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597#-L7ekvG6rOX18sJ414M6:-L82WVgqfR-qGL6xICEM:b-m2an5l) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/711a17bc5bc62f91783d4156e08703ca0b74348a/src/app/geo/identify.service.js#L97)):*

I really like this approach.  Maximum flexibility with minimal complexity on the API. Outstanding.
</blockquote></details>

Updated snippets to test if the identify is working correctly:
```ts
const t0 = performance.now();

// watching results as they come in
RZ.mapInstances[0].layers.identify.subscribe(event => {
    console.log('%c Parsing identify results as they resolve', 'background: #222; color: #bada55');

    event.requests.forEach((request, index) =>
        request.features.then(features =>
            console.log(index, request.layer, features, `resolved in ${(performance.now() - t0).toFixed(2)} milliseconds`)
        )
    );
});

// waiting for all results to clear
RZ.mapInstances[0].layers.identify.subscribe(event => {
    Promise.all(event.requests.map(request => request.features)).then(requestsFeatures => {
        console.log('%c Waiting for all identify results to resolve first', 'background: #222; color: #bada55');
        const t1 = performance.now() - t0;

        requestsFeatures.forEach((features, index) =>
            console.log(index, event.requests[index].layer, features, `resolved in ${t1.toFixed(2)} milliseconds`)
        );

        // we can still easily remove all but one results
        // requestsFeatures.forEach(features => features.splice(1));
    });
});
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597)*

milesap:
Just to save me a few minutes can you bump the version in `package.json` to `v2.3.0`.  I split #2601 into two parts and assigned the one to you. 

---

Review status: 2 of 5 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597#-:-L82ep1gyDZtzK087gMZ:b-wii84z)*

milesap:
1. Update `package.json` to `v2.3.0`
2. Confirm importing `RV` is ok (per @jahidahmed)
3. Confirm PR conforms to and closed https://github.com/fgpv-vpgf/fgpv-vpgf/issues/2601 (I believe so)

Then we should be ok to merge

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597#-:-L87Zguc1aHHUyVE8-vR:by3ga19)*

AleksueiR:
So, is it okay to import `RV` or not?

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/2597#-:-L87rTva1_nVw3D8xoIM:b13xwz)*

AleksueiR:
I was suggested to add `identifyMode` to each `ConfigLayer` object. Upon some pondering, I think doing this will add more complexity to the RAMP code with no clear benefits. Yes, it's flexible since the host page can configure some layer to output details, and some layers to just highlight identify results, but the RAMP's identify logic will need to be more complicated as well.

Right now, `identifyMode` is checked only once, when the identify is triggered. If it's to be present on all the layers, we need to write more checks to loop through all the layers and their respective modes. aggregate them and decide if the identify should be run at all, and then on which layers to run which identify mode, and which results to highlight, and so forth. 

We already have a config switch to turn the identify for a specific layer. Combined with the `identifyMode` they should be enough for majority of cases.

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/2597#-:-L87tUi9EsZSRt6T3X5H:bhpjieg)*

milesap:
Instead of importing `RV`, put

```js
export interface DataItem {
    name: string;
    value: string | number;
}
```

inside `layers.ts` and import that directly.

Your `identifyMode` has not yet been planned in the API spec. I'll review your proposal and let you know hopefully by the end of the day.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597#-:-L87x6sZ8Pz3LaM79jfg:b-thxj92)*

milesap:
:lgtm: 

---

Reviewed 1 of 3 files at r2, 2 of 2 files at r3, 2 of 2 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fgpv-vpgf/fgpv-vpgf/2597#-:-L88GB7evvzO20kuoioO:bg7igkv)*

PR #2889, opened 2018-07-27
Use custom angular material build

dane-thomas:
## Description

_New PR, missed a module_

Remove the angular-material dependency and include our own custom build. 
Reduces the size of included angular material to 656KB from ~1.2MB for the entire package.
This should improve slightly more after #2844 gets pulled since this benefits from css minification.
**Actual reduction**: 2506101 KB -> 2412487 KB

Added a script `generateAngularMaterial.js` that should make modifying the custom build simple. 
Modify the configuration, then run `npm run angular-material` and the build will be updated. 

## Testing

:eyes:
**_This needs thorough testing_**
I think I got all the modules we're using but some may have slipped through. 
The more :eyes: the better!

## Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/2886)


milesap:
Will change commit hashes and re-open

Issue #2942, opened 2018-08-17
Collapsed symbology stack is rendered with a black halo


AleksueiR:
### Problem
Symbology stacks with lots of items are rendered with a kind of a black halo when collapsed.

Demo url: http://fgpv.cloudapp.net/demo/develop/dev/samples/index-samples.html?keys=JOSM,CESI_Other,NPRI_CO

![](https://i.imgur.com/629PUXE.png)

### Expected result
![](https://i.imgur.com/73yH7aI.png)

YaroRoodzit:
Issue resolved. Tested with build  v2.5.0-b4.

PR #2958, opened 2018-08-27
fix(symbology): remove black halo from symbology stack

ShrutiVellanki:
## Description

Closes https://github.com/fgpv-vpgf/fgpv-vpgf/issues/2942

## Testing

http://fgpv.cloudapp.net/demo/users/ShrutiVellanki/halo-symbology/dev/samples/index-samples.html?keys=JOSM

## Documentation
Inline comments

## Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/2958)

### DIFF
--- a/src/app/ui/toc/templates/symbology-stack.html
+++ b/src/app/ui/toc/templates/symbology-stack.html
@@ -1,6 +1,6 @@
 <div
-    class="rv-symbology-container rv-render-{{ self.symbology.renderStyle }}"
-    ng-class='[{ "rv-expanded" : self.isExpanded }, { "rv-compound" : self.symbology.stack.length > 1 }]'>
+    class="rv-symbology-container rv-render-{{ self.symbology.renderStyle }} rv-compound"
+    ng-class='[{ "rv-expanded" : self.isExpanded }, { "rv-one" : self.symbology.stack.length === 1 }]'>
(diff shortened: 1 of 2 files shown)


ShrutiVellanki:
@AleksueiR 

So I seem to have found the root of the problem though I'm still confused...

in this section of the HTML: 
```html
<div class="rv-symbology-container rv-render-{{ self.symbology.renderStyle }}"
    ng-class='[{ "rv-expanded" : self.isExpanded, "rv-compound" : self.symbology.stack}]'>
<div class="rv-description-container">
```
the template binding is still working as far as `self.symbology.stack` returning the right value and applying `rv-compound` to the class list. However sometimes the classlist on the element returns `rv-compound-add` (which causes the issue):

capture

as opposed to this: (which keeps it working smoothly)

capture2

I thought this was strange since `rv-compound-add` is no where to be found in the code base. I could avoid this problem entirely by just doing this (doesn't effect how symbology stacks with one symbol look)

```html
<div class="rv-symbology-container rv-render-{{ self.symbology.renderStyle }} rv-compound"
    ng-class='[{ "rv-expanded" : self.isExpanded}]'>
<div class="rv-description-container">
```

Issue #2456, opened 2017-10-27
Improve symbology toggles behaviour


AleksueiR:
The behaviour of the symbology toggles is inconsistent with the layer group/child behaviour found elsewhere in the layer selector. Suggested improvements below:

| current                                      | improved                                 |
| ---------------------------------------- | ---------------------------------------- |
| Turning the layer "off" has no effect on symbology toggles. | Turning the layer "off", turn all the symbology toggles off. |
| ![ehn_symbology_toggles_1](https://user-images.githubusercontent.com/2285779/32109775-ae43f818-bb03-11e7-85f2-6650297c9cfd.gif) | ![ehn_symbology_toggles_2](https://user-images.githubusercontent.com/2285779/32109776-ae5961bc-bb03-11e7-8d90-b1a48dba9f60.gif) |
| Turning all symbology toggles "off" has no effect on the parent layer. | Turning all symbology toggles "off", turns the parent layer "off" as well. |
| ![ehn_symbology_toggles_3](https://user-images.githubusercontent.com/2285779/32109779-ae68e92a-bb03-11e7-9a54-8fffd85720c4.gif) | ![ehn_symbology_toggles_4](https://user-images.githubusercontent.com/2285779/32109780-ae78eda2-bb03-11e7-9b31-8810705d8e67.gif) |
| Turning a single symbology toggle "on" when all other toggles and the parent layer itself are "off" has no effect on the parent layer. | Turning a single symbology toggle "on" when all other toggles and the parent layer itself are "off" turns the parent layer back "on". |
| ![ehn_symbology_toggles_5](https://user-images.githubusercontent.com/2285779/32109781-ae8ce776-bb03-11e7-9e13-0e8b2148fffa.gif) | ![ehn_symbology_toggles_6](https://user-images.githubusercontent.com/2285779/32109774-ae31e056-bb03-11e7-84f8-747eb4bba359.gif) |

JahidAhmed:
Regarding the first one, based on current functionality, with the layer visibility off, we still see results in the table. However, if we modify it to toggle the symbology off when the layer is toggled off, the datatable will filter everything out and show no results whenever the layer is not visible. Is that behaviour fine?

AleksueiR:
That's a very good question. Maybe we can discuss this during our next RAMP meeting.

AleksueiR:
> Miles Petrov @milesap 13:19
I just had a pod discussion and think that toggling symbology off should mean:
[...]

PR #3051, opened 2018-09-28
fix(ui): toggle stack and layer behaviour in sync

ShrutiVellanki:
## Description

Closes https://github.com/fgpv-vpgf/fgpv-vpgf/issues/2456

## Testing

`keys=NPRI_CO, CESI_Other, JOSM`, `sample 46`, other samples with symbology stacks. 

**Test six cases:**
- Toggle off each item in stack, does top level layer toggle off automatically?
- Now toggle on an item in the stack, does top level layer toggle on automatically?
- Toggle off top level in the stack, do all items in the stack toggle off automatically?
- Toggle on top level in the stack, do all items in the stack toggle on automatically?
- Toggle off visibility of non-symbology stack layer, is the datatable empty?
- Toggle on visibility of non-symbology stack layer, is the datatable formatted properly?

Pay attention to correct entries showing up on datatable, and correct symbologies showing up on map.

## Documentation

inline comments

## Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3051)

### DIFF
--- a/src/app/geo/legend-block.class.js
+++ b/src/app/geo/legend-block.class.js
@@ -1,3 +1,5 @@
+import { Observable, Subject, BehaviorSubject } from 'rxjs';
+
 /**
  *
  * @module LegendBlock
@@ -658,6 +660,7 @@ function LegendBlockFactory(
             this.isControlDisabled = ref.isControlDisabled.bind(this);
             this.isControlSystemDisabled = ref.isControlSystemDisabled.bind(this);
             this.isControlUserDisabled = ref.isControlUserDisabled.bind(this);
+            this._selectedChanged = new Subject();
         }
 
         get isInteractive() {
@@ -673,6 +676,11 @@ function LegendBlockFactory(
         }
         set isSelected(value) {
             this._isSelected = value;
+            this._selectedChanged.next(value);
+        }
+
+        get selectedChanged(){
+            return this._selectedChanged.asObservable();
         }
 
         /**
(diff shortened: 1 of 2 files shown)


james-rae:
___
*[src/app/geo/legend-block.class.js, line 759 at r1](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3051#-LNWV2OH5xgIFpRvNP_N:-LNWujMQ7IGfSCEeaX-9:b-62brbv) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/c91a46aa18ca68e36611bfc692ada4162000e389/src/app/geo/legend-block.class.js#L759)):*
<details><summary><i>Previously, ShrutiVellanki (Shruti Vellanki) wrote…</i></summary><blockquote>

Updated comment. Applies visibility settings to grid upon creation. (by calling its `set visibility`, which calls `super` `set visibility`, which applies table filter)
</blockquote></details>

so spitball idea (you can squash it).  You put a function on `LegendEntry` called something like `updateSymbolFilter(visibility)`.  The contents are the same as `set visibility`.  We then call this function where it's needed instead of doing things like `visibility = visibility` or `super.visibility`.  

It won't change the logic, but when i see `.visibility = value` i think the code is storing something and changing it's visibility state, when actually it's doing a filter update.

JahidAhmed:
I was testing `No initial filter` on sample 46 (http://fgpv.cloudapp.net/demo/users/ShrutiVellanki/symbology-toggles-behaviour/dev/samples/index-samples.html?sample=46) and came across some weird behaviour.

If you toggle off two of the three items in the symbology stack, the layer is correctly displayed on the map. However, if you now open the datatable, the map will reset to showing all the items in the layer (disregarding the unchecked items in the stack) and the table also shows all entries (including unchecked items)

Issue #3079, opened 2018-10-11
Enhanced Table #2: Formatting + Rendering Issues and Enhancements


ShrutiVellanki:
List of improvements to be added to `enhancedTable` (mostly for the purpose of matching existing datagrid as closely as possible):

#### ~~1. Icon SVGs don't render properly~~ 
 ![3](https://user-images.githubusercontent.com/25359812/46810106-f2806500-cd3d-11e8-99d0-dae49851c4f5.PNG)
- ~~pictures don't show up~~
- ~~svg column doesn't show up at all the first time table open (shows up second time onward)~~
- ~~svg column should not have a header and be the first column (as per current datatables)~~

#### 2. ~~Titles are not formatted properly~~
  - `enhancedTable`:
![4a](https://user-images.githubusercontent.com/25359812/46810211-23609a00-cd3e-11e8-807d-279b342b8903.PNG)
- ~~current table (what we want):~~ 
![4b](https://user-images.githubusercontent.com/25359812/46810223-28bde480-cd3e-11e8-8707-b99012a95771.PNG)
- ~~also add tooltips on hover for titles that get cut off~~
#### ~~3. enhancedTable renders on top of simple table~~
  - ~~**Steps to reproduce:** open `enhancedTable`, close `enhancedTable`, you will see the simple table underneath which you have to close~~
#### 4. ~~Make sorts more conspicuous:~~
- ~~To a new user no way of knowing that clicking these means sorting:~~
![capture](https://user-images.githubusercontent.com/25359812/46813775-55292f00-cd45-11e8-8521-3df930fc512c.PNG)
- ~~up/down arrow only appears when clicked~~
- ~~want something like this which is present at all times:~~
![capture](https://user-images.githubusercontent.com/25359812/46813887-8275dd00-cd45-11e8-86c2-d545668a7a48.PNG)
#### 5. ~~No hovertext for close button (remember to add proper translation):~~ 
![capture](https://user-images.githubusercontent.com/25359812/46814038-ded8fc80-cd45-11e8-9178-c4264b46b577.PNG)
#### ~~6. Use material theme~~
- ~~Change the theme to material to fit with the rest of the app~~

PR #3092, opened 2018-10-16
fix(enhancedTable): close button, svgs and titles render properly

ShrutiVellanki:
## Description

Takes care of points 1, 2 and 5 in this issue: https://github.com/fgpv-vpgf/fgpv-vpgf/issues/3079
Changes in extensions repo: https://github.com/fgpv-vpgf/extensions/pull/6

## Testing

**[Test Link](http://fgpv.cloudapp.net/demo/users/ShrutiVellanki/rendering-fixes/dev/samples/index-enhanced-table.html)**

`enhancedTable` should now have:
- hovertip for close button
- properly formatted titles
- svgs that show up the first time table is opened
- properly rendered SVGs 
- tooltips that show up for titles that are cut off

## Documentation

inline comments

## Checklist

- ~[ ] Release notes have been updated~

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3092)


james-rae:
___
*[api/src/layers.ts, line 457 at r1](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3092#-LP17FFu4lKr9B5PGF4S:-LP1IiekEnEvzXi_-Adp:b-9mo6x9) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/8c5af5890d50cd69a4c7a7a6b1593c465d2c1b54/api/src/layers.ts#L457)):*
<details><summary><i>Previously, milesap (Miles Petrov) wrote…</i></summary><blockquote>

I'm thinking something like this would be better on an `attribute` object that gets pushed instead of the current attribute array. The object would be something like:

```js
{
  name: {
    id: 'STN_NAME',
    alias: 'Station Name'
  },
  values: [attribute rows...]
}
```

Lets pass this by james as well. 
</blockquote></details>

yes and no.   in the code further down in the file, we can see this private var is being populated by the appropriate attribute metadata (`field.alias`, `field.name`).   However, there is an additional property from the config, `layer.columns[].title`, which would override the other two.  This might be used already within the fancy table intention file (i don't see it in the minified diff in this PR, but it might be there).

Given we have a 3-level defaulting scheme for column titles, would suggest one of the following.
a) keep something similar to what we have, and add in the config `title` property to the logic.  this acts as a cache for the defaulting, making the grid creation code somewhat simpler and marginally faster
b) push all the column derivation logic to the grid generation routine.  this might be advantageous if we want to support being able to change things on the fly.  E.g. `API.changeAlias('my new alias')`, if we had option a this function would need to know to update `_formattedHeaders` which isn't obvious.  But option b the change would be auto-consumed the next time the grid opens.

PR #3166, opened 2018-12-03
feat(enhancedTable): add number and date filter types

### LINKED ISSUE #3121: Enhanced Table #13: Date Picker for Filter
### Problem
Ag-grid comes with support for date filters using the browser's built-in date picker. Since the implementation differs across browsers there's no guarantee for consistency in how it would work. Also the accessibility isn't great in Firefox and haven't been able to test it in IE. 

### Ideas
1. Create a controller for the date picker
  Do something similar to how the other buttons on the panel are made using angular. The main issue I see is the same controller would be used for multiple date pickers meaning the date values would be shared. Would appreciate if someone who knows more about how this works could confirm/deny.

2. Somehow compile the template that's being injected before adding it to the DOM
  Not sure if this is possible but should allow the use of the `md-datepicker`. 

3. Add a date picker to the panel api
  Would include the date picker as a panel element. Not sure how/if injecting this into the table and getting the values from it to the filter would work. 

4. Find a new date picker library
  Would need to find one that's both accessible and has a material theme. 

5. ~~Find a new grid?~~

My current preferred idea is 4 unless either 1 or 2 are possible. 

Any thoughts on the above or other ideas?

### LINKED ISSUE #3160: Enhanced Table #4: Filter By Numerical Data and add Regex Support
- Filtering by numerical data - i.e OID, latitude/longitude  - uses text matching in `enhancedTable` instead of the current functionality pictured below:  

- Filtering by date with accessible date picker

- add regex support as per current datagrid (`lazyFilter` which is the current default should only be applied when set through the config)

dane-thomas:
## Description

Supporting PR for https://github.com/fgpv-vpgf/plugins/pull/19

Closes #3121
Closes #3160 

Adds ability to compile angular with scope to the API. 
Exposes `tableNode` on the API so extensions can access table information

## Testing

:eyes:
http://fgpv.cloudapp.net/demo/users/dane-thomas/filter-types/prod/samples/index-enhanced-table.html

### Checklist

- [x] works in IE
- [ ] works via config file
- [ ] works via wizard
- [ ] works via API
- [ ] works via RCS
- [ ] works via bookmark load
- [ ] works in auto-legend
- [ ] works on layer reload
- [ ] datagrid works
- [ ] identify works

## Checklist

- [x] orignal issue has been reviewed & updated to reflect the PR content

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3166)


src/app/geo/init-map.directive.js
```diff
@@ -99,6 +99,15 @@ function rvInitMap($rootScope, configService, geoService, events, referenceServi
                 return $compile(html)($rootScope);
             };
 
+            // reference to the angular compiler that also returns the scope
+            apiMap.$compileWithScope = function(html) {
+                const scope = $rootScope.$new(true)
+                return {
+                    html: $compile(html)(scope),
```
milesap: Does the html need to be returned here? It was provided to the function so the caller should have a reference to it.

src/app/geo/init-map.directive.js
```diff
@@ -99,6 +99,15 @@ function rvInitMap($rootScope, configService, geoService, events, referenceServi
                 return $compile(html)($rootScope);
             };
```
milesap: Do we need two compile methods? I'm ok if we remove the above `$` compile function in favor of yours. If so please update plugins.

dane-thomas: I wasn't sure of all the locations using the compile function without scope but it's probably ok to change it.
Maybe keep it as `$` and add a flag for whether or not to use scope so existing functionality isn't changed?

src/app/geo/init-map.directive.js
dane-thomas: I thought `$compile` returns the new html after the input's compiled through angular so we would need the html returned there.

milesap: > Calling the linking function returns the element of the template. It is either the original element passed in, or the clone of the element if the cloneAttachFn is provided.
https://docs.angularjs.org/api/ng/service/$compile

Since `cloneAttachFn` is not passed, the original element passed in is altered.

src/app/geo/init-map.directive.js
milesap: Hmm ok, merge both into one and name it `$compile`.

```js
apiMap.$compile = function(html: JQuery<HTMLElement>, useIsolatedScope: boolean = true) {
    const scope = $rootScope.$new(useIsolatedScope);
    $compile(html)(scope);
    return scope;
}
``` 
It's relatively new and not widely used yet. 

Places that need an update
https://github.com/fgpv-vpgf/fgpv-vpgf/blob/v3/api/src/panel.ts#L195
https://github.com/fgpv-vpgf/fgpv-vpgf/blob/v3/api/src/panel.ts#L266
https://github.com/fgpv-vpgf/fgpv-vpgf/blob/v3/docs/developer/plugins.md

Instead of this

```js
this.elementAttr = this.panel.map.$($(element)[0]);
```

Do this

```js
this.elementAttr = $(element); // normalizes to a JQuery<HTMLElement>
this.panel.map.$compile(element[0]); // compiles the native DOM element in the above jQuery object.
```

PR #3210, opened 2019-01-04
fix(ie detection): update isIE to isIE11 and no longer include edge

### LINKED ISSUE #3165: Update IE detector to distinguish between IE11 and Edge
The existing IE detection logic lumps both IE and Edge together. Some functionality is disabled/modified for IE11, but it should still work in Edge as expected.

https://github.com/fgpv-vpgf/fgpv-vpgf/blob/develop/src/app/bootstrap.js#L13

Improve detection logic to only catch IE11.

milesap:
## Description
Provides an `isIE11` boolean flag on `appInfo` and the api which is true iff the browser is IE11. Previously is IE was true for both IE and Edge browsers.

Closes #3165

## Testing
👀 on chrome, IE11, and Edge

### Checklist

- [x] works in IE11
- [x] works in Edge
- [ ] works via config file
- [ ] works via wizard
- [x] works via API
- [ ] works via RCS
- [ ] works via bookmark load
- [ ] works in auto-legend
- [ ] works on layer reload
- [ ] datagrid works
- [ ] identify works

## Documentation
None

## Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3210)


milesap:
@AleksueiR Right, it's too early to use `appInfo` in that case.

Issue #3244, opened 2019-01-22
Build errors on v3 branch


milesap:
Investigate builds failing for `develop` and `prod` environments. Node is throwing "Module not found" errors for select geoApi dependencies. Check:

1. geoApi lists all it's imported dependencies in package.json
2. webpack continues to include the geoApi `node_modules` folder in the build.

```
ERROR in ./node_modules/geoApi/node_modules/shpjs/lib/unzip.js
Module not found: Error: Can't resolve 'jszip' in '/home/milesap/Documents/meeseeks-vm/tmp/repos/aa/node_modules/geoApi/node_modules/shpjs/lib'
 @ ./node_modules/geoApi/node_modules/shpjs/lib/unzip.js 3:12-28
 @ ./node_modules/geoApi/node_modules/shpjs/lib/index.js
 @ ./node_modules/geoApi/src/layer.js
 @ ./node_modules/geoApi/src/index.js
 @ ./src/app/core/config.service.js
 @ ./src/app/core/core-loader.js
 @ ./src/app/app-loader.js

ERROR in ./node_modules/geoApi/node_modules/shpjs/lib/index.js
Module not found: Error: Can't resolve 'lie' in '/home/milesap/Documents/meeseeks-vm/tmp/repos/aa/node_modules/geoApi/node_modules/shpjs/lib'
 @ ./node_modules/geoApi/node_modules/shpjs/lib/index.js 8:14-28
 @ ./node_modules/geoApi/src/layer.js
 @ ./node_modules/geoApi/src/index.js
 @ ./src/app/core/config.service.js
 @ ./src/app/core/core-loader.js
 @ ./src/app/app-loader.js

ERROR in ./node_modules/geoApi/node_modules/shpjs/lib/binaryajax-browser.js
Module not found: Error: Can't resolve 'lie' in '/home/milesap/Documents/meeseeks-vm/tmp/repos/aa/node_modules/geoApi/node_modules/shpjs/lib'
 @ ./node_modules/geoApi/node_modules/shpjs/lib/binaryajax-browser.js 3:14-28
 @ ./node_modules/geoApi/node_modules/shpjs/lib/index.js
 @ ./node_modules/geoApi/src/layer.js
 @ ./node_modules/geoApi/src/index.js
 @ ./src/app/core/config.service.js
 @ ./src/app/core/core-loader.js
 @ ./src/app/app-loader.js

[...]

PR #3247, opened 2019-01-23
Add non proxied legacy api as _RV to plugin instances + user checks

### LINKED ISSUE #3246: Plugins: scope legacy api
Plugins which use the legacy api (like back-to-cart) need the `legacy-api.js` file to be loaded before ramp. Some of these plugins will also be bundled with ramp on v3 release and would fail if a user does not include the `legacy-api.js` file. 

I suggest we add the non-proxied legacy api on the plugin prototype as `_RV`. 

Inside the plugin it is accessible everywhere once `init` is called.

```js
init(api) {
  this.doSomething()
}

doSomething() {
  this._RV.getCurrentLang(); // returns 'en-CA'
}
```

Not officially supported.

milesap:
## Description
Plugin instances have access to a non proxied version of the legacy api via the `_RV` property - available once the `init` method is invoked:

```js
const myPlugin = {
  init: function(api) {
    console.log(this._RV.getCurrentLang()); // 'en-CA'
  }
};
```

The `LEGACY_API` constant within ramp core is now also a non proxied version of the legacy api and is available regardless of `legacy_api.js` being present on the host page.

Added a console message when accessing the proxied api that it is deprecated.

Added a check to the plugin loader that prevents ramp from crashing when a plugin is registered on `rv-plugins` but is not defined.

Closes #3246, #3244

## Testing
Chrome, Edge 15. Tested with demo plugin.

### Checklist

- [ ] works in IE11
- [x] works in Edge
- [ ] works via config file
- [ ] works via wizard
- [x] works via API
- [ ] works via RCS
- [ ] works via bookmark load
- [ ] works in auto-legend
- [ ] works on layer reload
- [ ] datagrid works
- [ ] identify works

## Documentation
`_RV` is not officially supported and is intended for our own plugins - particularly pre-bundled ones. 

## Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3247)

### DIFF
--- a/api/plugin-loader.ts
+++ b/api/plugin-loader.ts
@@ -41,6 +41,11 @@ export default class Loader {
             .forEach((p: any) => {
                 const plugin = this.initialize((<any>window)[p]);
 
+                if (!plugin) {
+                    console.error(`Plugin with name ${p} is not defined at window.${p}. This plugin is ignored.`);
+                    return;
+                }
+
                 if (plugin.intention && loadIntentions[plugin.intention]) {
                     delete loadIntentions[plugin.intention];
                     this.intentions[plugin.intention] = plugin;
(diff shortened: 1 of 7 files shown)


milesap:
Pushed updated `package-lock.json` which fixes build errors (issue #3244).

Issue #3156, opened 2018-11-29
Enhanced Table #8: Apply to Map button


dane-thomas:
Branches off from #3088

In the mentioned issue there was an idea to scrap the settings panel on the old table and move all functionality from it to the table itself. Additionally with some of the design changes that were decided for the table, there is no longer a footer where we can put an `APPLY` button to apply the filters to the map. 

There are currently three ideas for migrating this functionality to the new table:
1. **Debounce the input to the filters and apply directly to map**
    Modify the filters so they wait for a specified amount of time then apply the filter from the column to the layer with out any button press from the user. 

2. **Button in the header**
    Add a button in the header that would apply the filters to the map. It would need to be decided where the button would go and how to place it since the header is somewhat crowded already. 

3. **Apply button for column filters**
    For ag-grid there's an option to add a button to apply the column filter. We could use this same button to trigger the apply to map. 

For now, the preferred solution is option 1 since it prevents us from having to add another button to the table panel and mimics the way the filters currently work for the table.

mweech:
Option #1 sounds like what I would expect to happen on filtering the grid, but may not be expected functionality for all so we can not just switch without having a fall back to a "lazy apply".  How about a toggle option in the menu?

mweech:
@dane-thomas can you please summarize why this is blocked.

james-rae:
It's waiting for me & the filtering engine ( https://github.com/fgpv-vpgf/fgpv-vpgf/issues/3055 ).  Once my code is working the apply to map trigger should be able to plug in, provide it's info, and everything else updates itself.

dane-thomas:
After discussion with @AleksueiR we decided to change the `Apply to map` button to `refresh map data`. 
![image](https://user-images.githubusercontent.com/8885149/51039874-3759a900-1584-11e9-8a05-6d880952e5af.png)

The button has been added to the UI with a placeholder function that needs to be filled in once the filtering engine is complete. 
[Code Changes](https://github.com/dane-thomas/plugins/commit/edd8e3de43d0bcf69ce53b507a6aefeafcca480c)

PR #3261, opened 2019-01-30
feat(enhancedTable): apply to map button and filter integration

dane-thomas:
## Description

#3156 

## Testing

:eyes:

### Checklist

- [x] works in IE11
- [x] works in Edge
- [ ] works via config file
- [ ] works via wizard
- [ ] works via API
- [ ] works via RCS
- [ ] works via bookmark load
- [ ] works in auto-legend
- [ ] works on layer reload
- [x] datagrid works
- [ ] identify works

## Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3261)


ShrutiVellanki:
Overall, excellent. :tada: 

So few points, (you did mention to me you were aware of problems with symbol toggles. If some of these are caused by the same thing, feel free to ignore.): 


 **Problem 1: Grey out apply to map button when no filters are present**
image


 **Problem 2: Asphalt Refinery**

1.Toggle off all layers expect Asphalt Refinery, you will get two points on the map: 
image


2. Open table and press apply to map, you will have 8 points: 
image


is there a reason for this behaviour? I understand applying filters and going from more points to less points, but not the other way around.

 **Problem 3: Non Consistent Behaviour with Global Search**

1. Toggle all layers off except  `Has initial filters`
2. Open table and apply to map:
- you will only get one point correspond to `Ener`
image

3. Now toggle on `3 columns no filters`, type mexico into global search. 
4. Type 1 in OID min search and apply to map. Map shows non-mexico points.

image

5. Type in 5 in OID max search and apply to map. Now map shows only mexico points: 
image

dane-thomas:
*Problem 1:*
You don't actually want this since you might want to apply no filters to the layer. For example if you applied a column filter to the map and then wanted to clear that from the map you'd need to apply the empty filter. 

**Problem 2:**
Not sure about this one. It appears that the layer is being filtered and that filter isn't being passed to the table so applying to the map overrides the filter.

**Problem 3:**
Nice catch. `mexico` wasn't matching `Mexico` in my checks but was matching in the table!
Will be fixed in an upcoming PR

james-rae:
Looking at problem 2 post-refactor.  The layer has an initial grid filter, filtering to object ids in range 0-20. There are only two features that match this.  The layer is filtering correctly, and it does this by having a `GRID` sql filter on the layers filter object.  When the grid opens, it LOOKS like it's not respecting the initial filter in the config.  Im not that familiar with how this worked in v2....if the grid is supposed to read from the config file `columns` array and pre-init it's filters, then might be a bug there.  Anway something to investigate, will not be fixed on refactor PR.

PR #3346, opened 2019-03-11
fix(enhancedTable): ui fixes

### LINKED ISSUE #3338: Enhanced table: UI fixes needed
Some small tweaks for the table panel:

## 1. Header is not delineated
Material panels use a light-grey border to separate their headers from the body. The enhanced table is missing that.

We also can move the number of records shown into the header as a subtitle:

## 2. The table padding is inconsistent
Horizontal padding of the panel is much larger then vertical:

The default padding value for panels is 8px. Try that first and see how it looks. The horizontal and vertical values might differ a bit, but the result should appear balanced. Something like this:

## 3. Header icon-button are incorrectly sized
The default size of the icon-buttons is 40x40 px (see the icons in the main app bar). The size of the icon itself is 24x24 plus the 8px padding. Right now, table header icon-buttons are 36x32 and look bunched up.

Also, please add a vertical line between the search icon and the column selector button, like in the main app bar:

## 4. The icon-button focus highlight is eccentric
When focused, the the icon and its ripple container are not centered (the ripple actually extends outside the button container):

The rippled container should be fully inside the button container and centered.

## 5. The focus highlight on column reorder button looks odd
It's a small icon-button, so the ripple container is a circle, but the button itself is squashed squeezing the ripple as well (the ripple is offset as well). 

I think, in this case, it would be better to use a rectangular ripple container, like one used on the text buttons - it will match nicely the hover/focus effect on the column header button:

dane-thomas:
## Description

Closes #3338

## Testing

Any sample page should work

### Checklist

- [x] works in IE11
- [x] works in Edge
- [ ] works via config file
- [ ] works via wizard
- [ ] works via API
- [ ] works via RCS
- [ ] works via bookmark load
- [ ] works in auto-legend
- [ ] works on layer reload
- [x] datagrid works
- [ ] identify works

## Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3346)


dane-thomas:
@AleksueiR @ShrutiVellanki yea it seems like icon buttons have a round ripple by default and there doesn't seem to be a simple way to make it rectangular

Issue #3294, opened 2019-02-14
Enhanced Table: keyboard accessibility issues between filters and grid


dane-thomas:
`ensureDomOrder` needs to be set to true to make the table accessible for screen readers. 

This was enabled in https://github.com/fgpv-vpgf/plugins/pull/50 but doing so causes some strange behavior when using the keyboard to navigate the table:

**1**. Tabbing between the last column filter and the grid will sometimes shift the focus to the `<body>` but only if the grid had been focused using the mouse
**2**. Pressing the enter key to use the move column left button shifts the focus to the `<body>`
**3**. Pressing the home button after scrolling horizontally will move the first cell to look focused but will sometimes focus the `<body>`

dane-thomas:
Bad ideas: 

1. **Don't enable `ensureDomOrder`**
     Can't do this because it could cause issues with screen readers.

2. **Use Focus Manager**
    Create a link between the last column filter and the first cell of the table.
    Won't help for issues 2 and 3.
    Need to update link when filtering changes which cell is first and column hiding/showing changes the last column filter. 

3. **Use event listener**
    Essentially same as above but using event listeners instead of focus manager links. 
    Will run into the same issues as stated above

4. **setTimeout** 
    A _temporary solution_ if we need to get this working soon.
    Have a delay long enough that the focus fails and switches to the `<body>` then force the focus change. 
    The downside to this is race conditions and a proper solution would need to be determined rather quickly.

milesap:
I tried to replicate the above issues while the focus manager is disabled and it seems everything is resolved. Please try this as well by commenting out this line https://github.com/fgpv-vpgf/fgpv-vpgf/blob/3df35dc3f7f73c059b2acd1f6f2beee5f60375ab/src/app/geo/init-map.directive.js#L93

The FM looks for a parent containing the property `rv-focus-exempt` as an indication to ignore the current focus action, but when I set this property on the panel itself some of the issues persisted, indicating the FM is still operating in some capacity. It's likely the FM is acting on events such as focus loss/gain where it does not look for this property. The performance implications of this check on certain parts of the FM may make this a non-option.

Anyway, check if the FM is to blame and if so we can look further into disabling it for certain parts of the viewer/plugins.

See: https://github.com/fgpv-vpgf/fgpv-vpgf/blob/3df35dc3f7f73c059b2acd1f6f2beee5f60375ab/src/app/focus-manager.js#L470

[...]

PR #3418, opened 2019-04-12
fix(enhancedTable): enable accessible list navigation

ShrutiVellanki:
## Description

- Closes https://github.com/fgpv-vpgf/fgpv-vpgf/issues/3294
- [`prepEnhancedTable`](https://github.com/fgpv-vpgf/fgpv-vpgf/compare/master...ShrutiVellanki:accessibility-2?expand=1#diff-44f6d9cfee9dca5205c7134180d52558R344) belongs in plugins. I'll move it as soon as I can get my `node` to start working again so I can test properly :sleepy: 
- cell navigation is broken see: https://github.com/fgpv-vpgf/fgpv-vpgf/issues/3414
- also i can't figure out a way to get the scrollbar to scroll when i'm navigating using arrow keys 
(so it'll sometimes appear like nothing is being highlighted because the scrollbar will be all the way to the right but keep pressing `downarrow` and you should see the highlights), suggestions are welcome

## [Testing](http://fgpv.cloudapp.net/demo/users/ShrutiVellanki/accessibility-2/dev/samples/index-samples.html?sample=2)

- tab for list navigation, up and down arrows for item navigation, tab again for object navigation
- should highlight for items and have a border for lists
- should focus out to the parent element and remove any item highlights when tabbing is finished

### Checklist

- [ ] works in IE11
- [ ] works in Edge
- [ ] works with projection change
- [ ] works with language change
- [ ] works via config file
- [ ] works via wizard
- [ ] works via API
- [ ] works via RCS
- [ ] works via bookmark load
- [ ] works in auto-legend
- [ ] works in structured legend
- [ ] works on layer reload
- [x] datagrid works
- [ ] identify works

## Documentation

inline comments

## Checklist

- [x] Commit messages follow [the guidelines](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/master/CONTRIBUTING.md#-git-commit-guidelines)
- [ ] Release notes have been updated
- [x] PR targets the correct release version
- ~[ ] Help files and documentation have been updated~
- [x] original issue has been reviewed & updated to reflect the PR content

Remember, it is a *muffin offence* to open a PR with any of the above checklist items incomplete.

Please keep the original issue up to date with the final solution, expected behaviour, and any additional notes for testers

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3418)

Issue #3414, opened 2019-04-11
EnhancedTable: broken cell navigation


ShrutiVellanki:
http://fgpv.cloudapp.net/demo/v3-panel-migration/dev/samples/index-samples.html

Try navigating through cells of table by tabbing or using arrow keys, navigation does not work properly. 
- This should work properly as a part of this: https://github.com/fgpv-vpgf/fgpv-vpgf/issues/3294 larger issue
- However I will not be overriding with my new `InternalsFocusManager` it on @AleksueiR suggestion because it would be more ideal to restore the `ag-grid` tabbing system as it worked before 

### Points:

- Clicking into a cell sometimes highlights and sometimes does not. The cell does get selected as using the arrow keys after seems to move the focus to the cells around them. When it does highlight (in chrome) it gets stuck highlighting the clicked cell and also highlights the cell being moved to. 

- Clicking or using the keyboard to activate the details or zoom buttons works but the cell doesn't stay highlighted although it is still focused. 

- Tabbing from the filter lists to the grid without highlighting any of the filters doesn't focus a cell in the table.
- Reaching either the left or right edge of the visible table when navigating using the arrow keys causes the arrows to switch to controlling the scrollbars for a short time then back to controlling cell selection.

ShrutiVellanki:
Fixed by this: https://github.com/fgpv-vpgf/fgpv-vpgf/pull/3421

YaroRoodzit:
Tested build: v3.0.0-b4
Issue solved.

PR #3421, opened 2019-04-15
fix(enhancedTable): enable accessible list navigation

ShrutiVellanki:
## Description

- Closes https://github.com/fgpv-vpgf/fgpv-vpgf/issues/3294
- cell navigation is broken see: https://github.com/fgpv-vpgf/fgpv-vpgf/issues/3414
- This is basically this: https://github.com/fgpv-vpgf/fgpv-vpgf/pull/3418 with a couple fixes added in:  https://github.com/fgpv-vpgf/fgpv-vpgf/pull/3418#issuecomment-483418517
- Sister PR: https://github.com/fgpv-vpgf/plugins/pull/78

## [Testing](http://fgpv.cloudapp.net/demo/users/ShrutiVellanki/accessibility-5/dev/samples/index-samples.html?sample=2)

- tab for list navigation, right and left arrows for item navigation, tab again for object navigation
- should highlight for items and have a border for lists
- should focus out to the parent element and remove any item highlights when tabbing is finished

### Checklist

- [ ] works in IE11
- [ ] works in Edge
- [ ] works via config file
- [ ] works via wizard
- [ ] works via API
- [ ] works via RCS
- [ ] works via bookmark load
- [ ] works in auto-legend
- [ ] works on layer reload
- [x] datagrid works
- [ ] identify works

## Documentation

inline comments

## Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3421)

### DIFF
--- a/api/src/panel/index.ts
+++ b/api/src/panel/index.ts
@@ -305,12 +305,14 @@ export class Panel {
         this.positionChangedSubject = new Subject();
         this.widthChangedSubject = new Subject();
         this.heightChangedSubject = new Subject();
+        this.listInit = new Subject();
 
         this.opening = this.openingSubject.asObservable();
         this.closing = this.closingSubject.asObservable();
         this.positionChanged = this.positionChangedSubject.asObservable();
         this.widthChanged = this.widthChangedSubject.asObservable();
         this.heightChanged = this.heightChangedSubject.asObservable();
+        this.populateList = this.listInit.asObservable();
     }
 
     /**
@@ -467,6 +469,7 @@ export interface Panel {
     positionChangedSubject: Subject<any>;
     widthChangedSubject: Subject<any>;
     heightChangedSubject: Subject<any>;
+    listInit: Subject<any>;
 
     //user accessible observables
     opening: Observable<Panel>;
@@ -474,6 +477,7 @@ export interface Panel {
     positionChanged: Observable<[number, number]>; //top left, bottom right
     widthChanged: Observable<number>;
     heightChanged: Observable<number>;
+    populateList: Observable<any>;
 
     appBar: appBar;
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -299,9 +299,8 @@
       "version": "github:claviska/jquery-minicolors#864e5991fadd77a9939c2350e44837193576422d"
     },
     "@fgpv/rv-plugins": {
-      "version": "3.0.0-3",
-      "resolved": "https://registry.npmjs.org/@fgpv/rv-plugins/-/rv-plugins-3.0.0-3.tgz",
-      "integrity": "sha512-4FJ6rxLuydAANcqgvlcwTncg5nOA0fW8q5QIZLLG7dlbbuw6AorrF+z039QHjFCreV0PzimbdE9fumsuTp2qEw==",
+      "version": "github:ShrutiVellanki/plugins#d8a8b816a88430e58c724a646f7babe69861c6e5",
+      "from": "github:ShrutiVellanki/plugins#d8a8b816a88430e58c724a646f7babe69861c6e5",
       "requires": {
         "ag-grid-community": "19.0.0",
         "rxjs": "6.3.3"
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "license": "MIT",
   "dependencies": {
     "@claviska/jquery-minicolors": "github:claviska/jquery-minicolors",
-    "@fgpv/rv-plugins": "3.0.0-3",
+    "@fgpv/rv-plugins": "github:ShrutiVellanki/plugins#d8a8b816a88430e58c724a646f7babe69861c6e5",
     "@flowjs/ng-flow": "2.7.8",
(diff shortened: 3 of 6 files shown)


AleksueiR:
Some unexpected behaviours:
- tab to a list and select an item with arrow keys
- press `tab` to focus on the first focusable node inside the list item
- observe that nothing happened

You need to press tab several times before any focusable nodes inside the item are highlighted. The number of time you need to press tab varies, but it seems it's related to the position of the item - the farther the item, the more you need to press tab.



---

- tab to a list and select an item with arrow keys
- press `tab` to focus on the first focusable node inside the list item
- press `up` or `down` key to select an adjacent item
- notice that the focus node in the previous list item is still focused (the actual focus should have gone back to the list container)
- press `tab` again
- notice that the item right after the list (the filter list) got focus instead of the first focusable node inside the selected list item



---

this one is weird: tabbing between the menu and panel close button slowly scrolls the body of the table to the right



---

- tab to a list and select an item with arrow keys
- focus on the first focusable node inside the list item
- press `shift+tab`
- notice that the item right after the list (the filter list) got focus instead of the focus moving back to the panel close button

spencerwahl:
Tabbing past the table to the nav buttons then hitting `shift + tab` sends you to the layers/legend button and not the table. I'm not sure if that's in scope though, since its an overall tab order issue and not internal to the table.

dane-thomas:
Also an issue when tabbing through the header. After tabbing through all the buttons, pressing tab again doesn't take you to the filters but skips straight to the nav buttons.

dane-thomas:
Clicking into a cell sometimes highlights and sometimes does not. The cell does get selected as using the arrow keys after seems to move the focus to the cells around them. When it does highlight (in chrome) it gets stuck highlighting the clicked cell and also highlights the cell being moved to. 

Clicking or using the keyboard to activate the details or zoom buttons works but the cell doesn't stay highlighted although it is still focused. 

Pressing `shift+tab` while moving through list items goes back to a seemingly random focusable element on the page and also keeps the highlight on the list item. 

Tabbing into the header list or filter list causes the table to scroll all the way left or all the way right. 

Tabbing from the filter lists to the grid without highlighting any of the filters doesn't focus a cell in the table.

ShrutiVellanki:
@dane-thomas summarizing our gitter conversation about table cells

For points 1, 2 and 5:
- I think cell stuff (for the most part) belongs here: https://github.com/fgpv-vpgf/fgpv-vpgf/issues/3414 so I'll move it there because your list will be useful for that issue

In this issue: 
- I'll try and make sure the grid takes you into the first cell (since that's a FM issue) and try to make sure that it always enters into the grid so that its navigable by arrows
- I'll also double check that apart from tabbing in and out of the "cell list" as a whole, my code isn't overriding default `ag-grid` behaviour

I'll address your other points in this issue as well.

dane-thomas:
There's a bit of lag when tabbing through certain parts of the table:
- close button -> header
- header -> filters
- filters -> grid

For some reason the reverse seems to work without any problems.

dane-thomas:
A few more things:
- Using the keyboard to change a  column position prevents any further keyboard navigation in Firefox.
- Sometimes moving left on the headers causes the highlighted header to snap left. Seems to only be in Firefox as well. Not sure how important this is but it would be somewhat jarring to a user. 
- Can't navigate through text of a filter. Using the arrow keys when typing into a filter moves to an adjacent filter rather than navigating through character. Again, not sure if this is an issue or expected behavior.

AleksueiR:
> There's a bit of lag when tabbing through certain parts of the table:

Yeah, I noticed that too. If it were FM, I would think the lag is from iterating though the DOM tree looking for focusables. Focused on the filter row, you can press the key twice or three times (if you do it quickly), and the focus will only move once.

When navigating using arrow keys, pressing and holding an arrow key only moves the selection once (only the adjacent item is highlighted). However, if you let go and press again, the selection moves to the last item in the direction of the key you were holding before (seems like internally the proper selection is tracked, but is not rendered).

ShrutiVellanki:
Resolved the following which should wrap up this PR: 

> ### TODO:
> * Using the keyboard to change a column position prevents any further keyboard navigation in Firefox.
> * Sometimes moving left on the headers causes the highlighted header to snap left. Seems to only be in Firefox as well. Not sure how important this is but it would be somewhat jarring to a user.
>   keyboard with an input added
> * `esc` shifts table weridly
> * when you use arrow keys to go through table cells and backtab, it shifts the header weirdly


### Remaining things: 
- > * bit of a lag when tabbing through certain parts of the table
- also there is a bit of a lag when using the keyboard to move a column to the left (when compared to moving to the right), this is because I am using [this code](https://github.com/fgpv-vpgf/fgpv-vpgf/pull/3421/files#diff-44f6d9cfee9dca5205c7134180d52558R401) to steal back focus from Focus Manager 
                                                         - as @AleksueiR pointed out the columns destroy themselves on move left, focus goes to the body, so I need to steal focus back and give it to `InternalFocusManager`

Issue #3472, opened 2019-05-09
Refactor index-samples 


dane-thomas:
Plan for splitting up the giant sample page into multiple smaller and focused pages. 

- [ ]  **index-layers**
    - [ ] Layer settings locked/diabled
        - Layer with opacity, visibility both disabled and locked
        - Layer with opacity disabled and locked
        - Layer with opacity locked
        - Layer with visibility disabled and locked
        - Layer with visibility disabled initially
        - Layer with visibility locked
        - Layer with Bounding box disabled
        - Layer with Query disabled
        - Layer with Snapshot disabled
        - Layer with Metadata disabled
        - Layer with Boundary zoom disabled
        - Layer with Reload disabled
        - Layer with Remove disabled
        - Layer with Settings disabled
        - Layer without option to view its data table
    - [ ] Layers
        - Layer Refresh Interval
        - Tile layer
        - Tile layer which is only valid in one of the basemap projections
        - WFS layer defined in config
        - All layer types
        - File based layers
    - [ ] Layer customizations
        - API 2 blocks for a single layer
        - Layer with metadata included
        - Details panel templating
        - Physical layer order must match config layer order
    - [ ] Dynamic
        - Dynamic layer with children split across multiple top level groups
        - Dynamic layer controls several of the children in the dynamic group
        - Dynamic layers with many different inital opacities
        - True Dynamic Layers with different image format
        - Dynamic layers display field configuration
    - [ ] Not true dynamic
        - Dynamic opacity controls for "not-true-dynamic" dynamic layers

- [ ] **index-legend**
    - [ ] Structure
        - Visibility sets
        - Legend entry with controlled layers
        - Group with Direct linking of dynamic child
        - Legend entry with one group linking to multiple layers in child entries
        - Nested visibility sets
        - Single entry collapsed set
        - Super group with tile layer, dynamic layer- Controls visibility/settings and disable visibility
        - Zoom scale dependent layer as a part of visibility group and standalone
        - Sample with hidden layer
        - Collapsed visibility set + hidden layer group
    - [ ] Entry format
        - Legend entry only Information section
        - Legend entry with symbology overridden by config file specified image file
        - Legend with layer import disabled
        - Legend entry with description beneath layer title
        - Big images in info section
        - Big images in symbology stack
        - Symbology stack expand by default
    - [ ] re-order disabled for auto legends
    - [ ] Layer Re-ordering enabled

- [ ] **index-map-customizations**
[...]

PR #3487, opened 2019-05-14
test(samples): layers sample page

RyanCoulsonCA:
## Description
Completes the `index-layers` component of https://github.com/fgpv-vpgf/fgpv-vpgf/issues/3472

### Checklist

- [x] works in IE11
- [x] works in Edge
- [x] works via config file
- [ ] works via wizard
- [ ] works via API
- [ ] works via RCS
- [ ] works via bookmark load
- [ ] works in auto-legend
- [ ] works on layer reload
- [ ] datagrid works
- [ ] identify works

## Checklist

- [ ] original issue has been reviewed & updated to reflect the PR content

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3487)

### DIFF
--- a/src/content/samples/new/config/config-sample-layers-01-controls.json
+++ b/src/content/samples/new/config/config-sample-layers-01-controls.json
@@ -1,0 +1,588 @@
+{
+  "ui": {
+    "title": "Custom Title",
+    "navBar": {
+      "zoom": "buttons",
+      "extra": [
+        "fullscreen",
+        "geoLocator",
+        "home",
+        "help"
+      ]
+    },
+    "sideMenu": {
+      "logo": true
+    },
+    "about": {
+      "folderName": "default"
+    },
+    "help": {
+      "folderName": "default"
+    },
+    "legend": {
+      "isOpen": {
+        "large": true,
+        "medium": true,
+        "small": false
+      }
+    }
+  },
+  "version": "2.0",
+  "language": "en",
+  "services": {
+    "proxyUrl": "http://167.40.64.74/wmsproxy/ws/wmsproxy/executeFromProxy",
+    "exportMapUrl": "http://section917.cloudapp.net/arcgis/rest/services/Utilities/PrintingTools/GPServer/Export%20Web%20Map%20Task",
+    "export": {
+      "title": {
+        "value": "Title"
+      },
+      "map": {},
+      "mapElements": {},
+      "legend": {},
+      "timeout": 5000,
+      "footnote": {
+        "value": "This is a footnote added from the configuration file. The note is very long so it should wrap on multiple lines when it reaches a certain limit in size. Maybe some user will want to use this as aplace holder to put a lot of information so we need to be able to wrap this content. Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere cubilia Curae; Fusce aliquet ante quis aliquet feugiat. Cras eget semper nunc, eu placerat purus. Nunc sed lacinia enim, ut sollicitudin quam. Nunc quis finibus massa, eget maximus enim. Donec ac nisl libero. Nunc eu pharetra arcu. Fusce luctus, magna cursus gravida tristique, risus nisi porttitor magna, ac dictum ipsum dui vel nulla. Integer id ornare augue. Quisque condimentum velit quis elementum porta. Sed dui enim, iaculis cursus diam volutpat, laoreet porta quam. Sed nec aliquet magna. Curabitur commodo fringilla metus, eu posuere sapien mollis nec."
+      }
+    },
+    "search": {
+      "serviceUrls": {
+        "geoNames": "http://geogratis.gc.ca/services/geoname/en/geonames.json",
+        "geoLocation": "http://geogratis.gc.ca/services/geolocation/en/locate?q=",
+        "geoSuggest": "http://geogratis.gc.ca/services/geolocation/en/suggest?q=",
+        "provinces": "http://geogratis.gc.ca/services/geoname/en/codes/province.json",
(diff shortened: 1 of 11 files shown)


RyanCoulsonCA:
> _[src/content/samples/new/config/config-sample-layers-01.json, line 92 at r1](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3487#-LewhjvO46o5JPJxEjTE:-LewhjvO46o5JPJxEjTF:bx1nygw) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/6c7eab0527797da74c0124f071c7c09395f64f3e/src/content/samples/new/config/config-sample-layers-01.json#L92)):_
> 
> > ```json
> >         "children": [
> >           {
> >             "layerId": "testLayer1"
> > ```
> 
> These tests are in a slightly different format than every other sample page, should be fine.

@yileifeng Yeah, I had to format this page slightly different due to the fact that structured legends don't allow you to delete layers (as far as I know), and I needed that for test 12. To solve this I had to an autopopulated legend instead, which doesn't allow me to use the same format as the other pages.

All of your other points should be fixed now, though.

james-rae:
For `index-layers`

The word `disabled` appears to be used for invisible/transparent in some cases, and "locked" in others (config 01 mostly).  To me, disabled and locked are the same thing.  Maybe changed disabled to be "invisible" and "0% opacity" (or something else if you think it flows better), and always use disable or locked (choose one) for a control that cant be interacted with.


Config 03. Because the first test is large in the legend, suggest either collapsing it as default, or re-arrange "Test 2" to be "Test 1" if you want a consistent look amongst the configs.
Same with config 04 (maybe push test 5 to test 1)

Config 04, test 3, we are probably good with one mid-opacity test. We can drop two of them (so keep zero, full, and one of the inbetweens)

Config 04, test 4.  maybe just a note to the QA that seeing an image draw on the map means things are working. They won't be inspecting file types on the web traffic.

Config 04, test 2. minor rewording. "controls" makes me think of some kind of controlled layer config. Maybe something like "Test 2: Layers with Children" and change "Power Plants Proper Subset" to "Layer with 3 Children"

Config 04, test 1. minor rewording. "children split" gives me impression they all belong to the same layer (its actually same service loaded 3 times). Maybe "Test 1: Layer Groups" and we can drop one of the layers (two is good).

Config 05, can we default-collapse all the test groups in the legend?

RyanCoulsonCA:
> > Reviewed 10 of 10 files at r3.
> > _[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3487)_ status: all files reviewed, 6 unresolved discussions (waiting on @dane-thomas, @RyanCoulsonCA, @ShrutiVellanki, and @yileifeng)
> 
> _[src/content/samples/new/config/config-sample-layers-01-options.json, line 1 at r3](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3487#-Lf051t878yunMmWjD0A:-Lf051t878yunMmWjD0B:b-wcumk) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/030c7ce0a8f6c379baa7d4f585902504ef8a1961/src/content/samples/new/config/config-sample-layers-01-options.json#L1)):_
> 
> > ```json
> > {
> > ```
> 
> Can you change the config name to match the name in the selector. So `options` -> `controls` maybe
> 
> _[src/content/samples/new/config/config-sample-layers-03.json, line 102 at r1](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3487#-Lewm8kG0WIzIqO56Yep:-Lf07Ei_4y8vhCGoJrDN:b-9si0l7) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/6c7eab0527797da74c0124f071c7c09395f64f3e/src/content/samples/new/config/config-sample-layers-03.json#L102)):_

Done.

> Previously, yileifeng (Yi Lei Feng) wrote…
> Some clarification about what this actually does:
> 
> All 4 legend blocks correspond to the same layer so any change to one of them should apply to the rest. It was primarily used to test removing legend entries through the legend API.
> 
> Could possibly not need to be in this sample at all.

I've removed it entirely from the sample for now.

RyanCoulsonCA:
> For `index-layers`
> 
> The word `disabled` appears to be used for invisible/transparent in some cases, and "locked" in others (config 01 mostly). To me, disabled and locked are the same thing. Maybe changed disabled to be "invisible" and "0% opacity" (or something else if you think it flows better), and always use disable or locked (choose one) for a control that cant be interacted with.
> 
> Config 03. Because the first test is large in the legend, suggest either collapsing it as default, or re-arrange "Test 2" to be "Test 1" if you want a consistent look amongst the configs.
> Same with config 04 (maybe push test 5 to test 1)
> 
> Config 04, test 3, we are probably good with one mid-opacity test. We can drop two of them (so keep zero, full, and one of the inbetweens)
> 
> Config 04, test 4. maybe just a note to the QA that seeing an image draw on the map means things are working. They won't be inspecting file types on the web traffic.
> 
> Config 04, test 2. minor rewording. "controls" makes me think of some kind of controlled layer config. Maybe something like "Test 2: Layers with Children" and change "Power Plants Proper Subset" to "Layer with 3 Children"
> 
> Config 04, test 1. minor rewording. "children split" gives me impression they all belong to the same layer (its actually same service loaded 3 times). Maybe "Test 1: Layer Groups" and we can drop one of the layers (two is good).
> 
> Config 05, can we default-collapse all the test groups in the legend?

These should be fixed.

Issue #3481, opened 2019-05-10
Initially, "Zoom to Feature"  does not work from Details window.


YaroRoodzit:
Tested build: v3.0.0-b2
"Zoom to Feature"  does not work from Details window or from details content window.
"Zoom to Feature" start to work properly only after use in datagrid. There is no problem with zooming used from data grid table. 
Issue exist for different browsers and can be seen with different layers.  
Steps to reproduce:
1. Use http://fgpv.cloudapp.net/demo/v3.0.0-b2/prod/samples/index-fgp-en.html?keys=JOSM
2. Click for details any map point 
3. Try clicking "Zoom to Feature" . Zoom failed. Error appears in console.
![image](https://user-images.githubusercontent.com/38046459/57549625-10449580-7332-11e9-8b15-53c00726e400.png)
4. Open content window and try "Zoom to Feature" . Zoom failed. Same error appears in console.
5. Close window and details. Open datagrid and click "Zoom to Feature". No issue. Working properly.
6. Close data grid . Click on back to Canada tab. Click any map point for details. Click "Zoom to Feature". No issue. Working properly.
7.  Open content window and try "Zoom to Feature". No issue. Working properly.

 Problem is back after page refresh or "back to Cart" used.

spencerwahl:
Zoom functionality was fixed to work with new panels. Additionally will fade out any panels covering the zoom area to ensure the point is visible.

YaroRoodzit:
Tested build: 3.0.0-b5
Issue resolved.

PR #3501, opened 2019-05-16
Fix for zoom offset calculation

spencerwahl:
## Description

#3481 

plugins PR: https://github.com/fgpv-vpgf/plugins/pull/95

Zoom wasn't working until the table set `externalOffset` because the calculation was using the old panels.
Now the calc uses the new panels PLUS it will fade any panel (from plugins, etc) that are on top of or very close to the zoom point.

## Testing

Details zoom should work in any sample without triggering a zoom from the table.

### Checklist

- [ ] works in IE11
- [ ] works in Edge
- [ ] works via config file
- [ ] works via wizard
- [ ] works via API
- [ ] works via RCS
- [ ] works via bookmark load
- [ ] works in auto-legend
- [ ] works on layer reload
- [x] datagrid works
- [x] identify works

## Documentation

Comments and function headers updated

## Checklist

- [] Release notes have been updated

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3501)

### DIFF
--- a/api/src/panel-registry.ts
+++ b/api/src/panel-registry.ts
@@ -124,4 +124,20 @@ export class PanelRegistry {
     get closing() {
         return this._panelClosing.asObservable();
     }
+
+    get panelOffset() {
+        // calculate what portion of the screen the main and filter panels take
+        const mainOpen = this.legend.isOpen || this.details.isOpen || this.fileLoader.isOpen || this.serviceLoader.isOpen || this.geoSearch.isOpen;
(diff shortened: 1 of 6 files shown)


JahidAhmed:
After zooming to a feature from the details panel and then panning a little and opening the table, it appears like this (semi-transparent):
image
It stays like that until the table is clicked on, at which point it goes back to normal. Wondering if that's expected behaviour

spencerwahl:
Ah, good catch... I forgot to filter on visible panels. I'll also try to make map interactions clear the peek.

Issue #3443, opened 2019-04-30
The viewer is stuck forever if shared URL is damaged.


YaroRoodzit:
Use "Get Link" or "Share" option to create URL . Delete or change part of the information string. Try to load modified URL. Viewer will be stuck forever on loading of  incomplete or damaged URL.

james-rae:
Adding some robustness to the bookmark parser is a good idea.
Will do in a future release

YaroRoodzit:
Tested build: v3.0.0-b5
Issue solved.

PR #3504, opened 2019-05-16
fix(bookmark): damaged URL does not break viewer

RyanCoulsonCA:
## Description
Fixes https://github.com/fgpv-vpgf/fgpv-vpgf/issues/3443

### Checklist

- [x] works in IE11
- [x] works in Edge
- [ ] works via config file
- [ ] works via wizard
- [ ] works via API
- [ ] works via RCS
- [ ] works via bookmark load
- [ ] works in auto-legend
- [ ] works on layer reload
- [ ] datagrid works
- [ ] identify works

## Checklist

- [ ] Commit messages follow [the guidelines](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/master/CONTRIBUTING.md#-git-commit-guidelines)
- [ ] PR targets the correct release version
- [ ] original issue has been reviewed & updated to reflect the PR content

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3504)

### DIFF
--- a/src/app/core/bookmark.service.js
+++ b/src/app/core/bookmark.service.js
@@ -389,14 +389,26 @@ function bookmarkService($q, configService, gapiService, bookmarkVersions, Geo, 
             }
         }
 
-        decodeMainBookmark();
+        try {
+            decodeMainBookmark();
+        } catch(err) {
+            // If the URL has been messed with, a couple errors could come up in this function.
+            console.error("An error occurred while trying to load the provided bookmark.",err);
+            return;
+        }
 
         // Make sure there are layers before trying to loop through them
         if (bookmarkObject.layers) {
             const layerData = bookmarkObject.layers.split(',');
 
             // create partial layer configs from layer bookmarks
-            bookmarkObject.bookmarkLayers = parseLayers(layerData, version);
+            try {
+                bookmarkObject.bookmarkLayers = parseLayers(layerData, version);
+            } catch(err) {
+                // Parsing layers may also fail if the URL is broken.
+                console.error("An error occurred while trying to load the provided bookmark.",err);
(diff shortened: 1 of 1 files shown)


RyanCoulsonCA:
For this bug, I found that the problem mainly occurred because of two things:  
1. atob() fails, if the URL cannot be decoded  
2. the URL can be decoded, but the decoded value is not the right type (i.e, the scale is a string)

My fix to this problem was basically to add some quick checks to see if either of these two problems occur. If either occurred, the config does not get updated. There's a chance I may have missed a case that causes the viewer to break, so if you find one let me know.

RyanCoulsonCA:
> for some reason `http://fgpv.cloudapp.net/demo/users/RyanCoulsonCA/fix-3443/prod/samples/index-fgp-en.html?rv=hello\world%\`
> 
> breaks it. Not sure if worth fixing?
> 
> 
> 
> > Reviewed 1 of 1 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3504#-:-Lf1OdjNA8oOC1JZ6oG2:bbvh244)_ status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dane-thomas and @yileifeng)

This looks like this may be an issue with the sample pages themselves. The page seems to crash if there's a % in a URL parameter. For example, this link also causes the same problem:
`http://fgpv.cloudapp.net/demo/users/RyanCoulsonCA/fix-3443/prod/samples/index-fgp-en.html?sdasdsad=%`

So I'm not sure if that problem should be fixed under this issue, or if a new one should be opened to deal with it.

PR #3519, opened 2019-05-23
fix(enhancedTable): a solution for overlapping arrows

### LINKED ISSUE #3507: Grid Column - Sort Arrow can overlap Reorder Arrow
Likely a combination of fixed width and long title.

AngryArrow

Sample to reproduce

```text
{
  "id": "AngryArrows",
  "name": "Electric Transmission Line",
  "layerType": "esriFeature",
  "table": {
    "columns": [
      {
        "data": "OBJECTID",
        "sort": "desc",
        "width": 200,
        "title": "This is a longer title for OBJECTID"
      },
      {
        "data": "Country",
        "width": 300
      }
    ]
  },
  "url": "http://geoappext.nrcan.gc.ca/arcgis/rest/services/NACEI/energy_infrastructure_of_north_america_en/MapServer/1"
}
```

yileifeng:
## Description
Fixes #3507 using CSS properties

## Testing
`config-sample-37.json` and `config-sample-38.json` are good tests for this, but test for different column widths and titles in the configs to ensure arrows never overlap. 

### Checklist

- [x] works in IE11
- [x] works in Edge
- [x] works via config file
- [ ] works via wizard
- [ ] works via API
- [ ] works via RCS
- [ ] works via bookmark load
- [ ] works in auto-legend
- [x] works on layer reload
- [x] datagrid works
- [ ] identify works

## Checklist

- [ ] PR targets the correct release version

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3519)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -300,8 +300,8 @@
       "from": "github:claviska/jquery-minicolors"
     },
     "@fgpv/rv-plugins": {
-      "version": "github:RyanCoulsonCA/plugins#0d700a30b563b5c741d06da23da6fd65926a7a0a",
-      "from": "github:RyanCoulsonCA/plugins#0d700a30b563b5c741d06da23da6fd65926a7a0a",
+      "version": "github:yileifeng/plugins#be9dd78e90988661b1e87d693361fc975c17e768",
+      "from": "github:yileifeng/plugins#be9dd78e90988661b1e87d693361fc975c17e768",
       "requires": {
         "ag-grid-community": "19.0.0",
         "rxjs": "6.3.3"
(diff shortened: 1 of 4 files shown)


dane-thomas:
I'm a little confused as to why these changes had to be handled viewer side. Since the table is a plugin I believe it would make more sense to fix the column widths there rather than internally modifying the values we're loading from the config. 

After a quick discussion with @JahidAhmed we weren't able to find a case where this would be the preferred way of handling it. Feel free to comment if you can think of any. 

One counter case would be if someone was to develop their own table that could handle smaller column widths than the enhanced table. It wouldn't make sense to them as to why their columns aren't getting as small as they wanted.

spencerwahl:
Were you able to try the CSS fix again after https://github.com/fgpv-vpgf/fgpv-vpgf/issues/3507#issuecomment-495279810

If so, what issues did you find?

yileifeng:
Still testing the CSS fix at the moment and it works well (the arrows never overlap anymore). It's just that when I use a width < 70px or so, the reorder arrows overlaps with the next column (maybe need to adjust padding or some other property).

ShrutiVellanki:
Since you are updating the `plugins` commit hash here, could you please make a PR with your changes on the plugins repo and link that PR here? 

Also I noticed that there is a certain "jumpy-ness" on your column headers. I think its because your reorder arrows are really close to the column title now, so they need to "move out of the way" to make room for the sort arrow: 

### Old table
jumpyHeaders

### New table

jumpyHead

spencerwahl:
To fix the issue @ShrutiVellanki brought up you can add a div between the `sort arrow` and `reorder arrows` with `flex: 1;` (it will also separate the `reorder arrows` and the `label` when there is no `sort arrow`)

Issue #3576, opened 2019-06-12
Failing to open datatable after changing languages 


yileifeng:
To replicate: 

Load
http://fgpv.cloudapp.net/demo/develop/dev/samples/index-fgp-en.html?keys=JOSM 

Change languages one time and try opening the datatable (it should be stuck loading). 

Change languages again and try opening the datatable (it does not load). Attempting to open the datatable in this case gives this error: 

![opening datatable changing languages issue](https://user-images.githubusercontent.com/31557789/59353806-cb759b00-8cf1-11e9-8c3c-fef7fd680346.PNG)

Changing languages in general will also cause a lot of additional errors but that is covered in #3534.

yileifeng:
Check again (should work?) if #3578 is merged

RyanCoulsonCA:
Just checked. Can confirm that the data table now opens correctly after a language change.

PR #3578, opened 2019-06-12
fix(*): fix console errors when changing languages

### LINKED ISSUE #3534: Changing the language causes a lot of errors in the console.
Tested build: v3.0.0-b5
When in IE a language change causes a lot of errors in the console.
The problem exists in other browsers.
image

yileifeng:
## Description
Closes #3534 

## Testing
Used `index-fgp-en.html?keys=JOSM` for testing. Changing languages should not produce the console errors mentioned in the original issue. However, if you test with a sample page from `index-samples`, then this error still appears when changing languages: 
index-samples dx error

### Checklist

- [x] works in IE11
- [x] works in Edge
- [x] works via config file
- ~~[ ] works via wizard~~
- ~~[ ] works via API~~
- ~~[ ] works via RCS~~
- ~~[ ] works via bookmark load~~
- [x] works in auto-legend
- [x] works on layer reload 
- [x] datagrid works
- [x] identify works

## Checklist

- ~~[ ] Release notes have been updated~~
- ~~[ ] Help files and documentation have been updated~~

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3578)

### DIFF
--- a/features/geosearch/geosearch.feature.ts
+++ b/features/geosearch/geosearch.feature.ts
@@ -136,13 +136,21 @@ class GeoSearchUI {
     }
 
     /**
-     * Retrun a list of formated province objects
+     * Return a list of formated province objects
      *
      * @return {Array} a list of formated province objects
      */
     fetchProvinces() {
         if (this.provinceList.length > 0) return this.provinceList; // in cache
         let provinceList = [];
+
+        const reset = {
+            code: -1,
+            abbr: '...',
+            name: '...'
+        };
+        provinceList.push(reset);
+
         let rawProvinces = (<any>this)._geoSearhObj.config.provinces.list;
         for (let code in rawProvinces) {
             provinceList.push({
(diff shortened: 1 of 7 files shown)


yileifeng:
In the process, this seems to have also fixed #3576 for me.

james-rae:
___
*[src/app/geo/legend.service.js, line 233 at r1](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3578#-LhBqLHe9X97h1HY8Mdp:-LhGVPmz5YvvF6zwHRIp:bs05q5h) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/35ac8d3762427db2a9f1147e1c9c91d12e4d6128/src/app/geo/legend.service.js#L233)):*
<details><summary><i>Previously, yileifeng (Yi Lei Feng) wrote…</i></summary><blockquote>

This was a bit strange, as the following error (caused by no `legendBlocks`)



only occurred on `index-fgp-en.html?keys=JOSM` for me, so i'm not sure on if its because of how that layer is setup differently or how the page is setup differently compared to a sample from `index-samples`. This occurs on language change (which seems to be the cause of a number of new issues popping up these days) and in this case, the layer that is being added after language change already exists from when the page first gets initialized. I'm not sure if there is another case besides language change where this error appears because of `legendBlocks` being undefined (I did not come across any) and if there is, that might be a concern. 
</blockquote></details>

ok.  sounds like it attemps to add too early, but then something else is adding it again later.  good for this PR

yileifeng:
Right, forgot to mention I will delete the commented out code if everything looks good

PR #3594, opened 2019-06-14
fix(core): prevent duplicate configService listeners

### LINKED ISSUE #3582: Duplicate listeners possible on config load
In `config.service.js`, the `onEveryConfigLoad (listener)` currently pushes duplicate listeners (where listener is a function to be triggered on config changes) to keep track when changing back and forth between languages. 

This can be observed by printing out the listeners when `setLang (lang)` gets called and identical functions will appear as elements in the list after changing the language at least twice. 

These duplicate listener functions do not seem to get properly de-registered which may or may not be the solution here. 

This was originally mentioned by @james-rae in https://github.com/fgpv-vpgf/fgpv-vpgf/pull/3578#pullrequestreview-248929412 and that PR currently adds a check to prevent duplicate listeners from being added.

yileifeng:
## Description
Closes #3582 

### Checklist

- [x] works in IE11
- [x] works in Edge
- [x] works via config file
- ~~[ ] works via wizard~~
- ~~[ ] works via API~~
- ~~[ ] works via RCS~~
- ~~[ ] works via bookmark load~~
- [x] works in auto-legend
- [x] works on layer reload
- [x] datagrid works
- [x] identify works

## Checklist

- ~~[ ] Release notes have been updated~~
- ~~[ ] Help files and documentation have been updated~~

### DIFF
--- a/src/app/core/config.service.js
+++ b/src/app/core/config.service.js
@@ -295,17 +295,20 @@ function configService($q, $rootElement, $http, $translate, events, gapiService,
             if (_loadingState >= States.LOADED) {
                 listener(getConfigByLanguage(currentLang()).config);
             }
-            // check for any duplicate listeners
-            if (!this.listeners.map(l => l.toString()).includes(listener.toString())) {
-                this.listeners.push(listener);
-            }
-            return () => {
-                const idx = this.listeners.indexOf(listener);
+
+            // check for any duplicate listeners and de-register if found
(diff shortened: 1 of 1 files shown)


dan-bowerman:
This is one that got lost along the way. It might be more effort than it's worth to figure this one out, and we're surviving fine with out it. Moving this to the "Blocked" column in ZenHub, but it's here for reference if this bubbles back up.

PR #3609, opened 2019-06-20
feat(identify): add option to trigger via the API

### LINKED ISSUE #3545: Add ability to perform identify query via API
Requested for CCCS Climate Data Viewer (https://climate-viewer.canada.ca).  

When the time period or season of the map is changed, the map layer needs to be refreshed.  This clears the current identify query and the user needs to re-select the grid point to see a value across multiple time periods.  This functionality would allow for the time period of the map to be changed so that the identify results can be updated as the underlying map layer is changed.

In order to facilitate this functionality, the map should allow an identify query to be initiated via API at a pre-defined set of coordinates.  This way when a control attempts to change the time period via a map layer change, it can have the identify query re-applied programatically.

#### Implementation Steps
- Continue to handle mouse click identify requests within the `identify` method in `identify.service.js` the same to prevent a breaking change on the API.
- Identify requests via the API will not emit mouse click streams. 
- All identify requests (mouse, api) will emit to a new observable on the map instance named `identified`.
- Identify panel/highlight will listen on the new observable.

#### Acceptance Criteria
1. Mouse based identify remains unchanged
2. Identify can be triggered via the API by providing coordinates
3. A new observable is created which emits on all identify types (mouse, api)
4. Both mouse and api identify requests open the identify panel and highlight the feature on the map (if enabled in config)

milesap:
## Description
Additions:
- `identify` method on API map instance accepts `XY` or `XYLiteral` coordinates
- `identified` observable on API map instance which emits whenever an API or mouse/keyboard identify is initiated. The emit value contains the identify coordinates (in lat/lon 4326) and a feature observable

Fixes:
- API geometry `XY` `projectToPoint` and `projectFromPoint` returns ESRI approved spatial references
- Undefined object error in symbology stack when identify was executed

Notes:
- No breaking changes to existing mouse/keyboard identify operations - click emits still occur in addition to the `identified` emit.
- Does this need to be documented in dev docs?

Closes #3545 

## Testing
```js
RAMP.mapInstances[0].identify([-79.347015, 42.651070]);
RAMP.mapInstances[0].identified.subscribe(console.log);
RAMP.mapInstances[0].identified.subscribe(x => x.features.subscribe(console.log));
```

### Checklist

- [x] works in IE11
- [x] works in Edge
- [x] works via config file
- [x] works via wizard
- [x] works via API
- [x] works via RCS
- [x] works via bookmark load
- [x] works in auto-legend
- [x] works on layer reload
- [x] datagrid works
- [x] identify works

## Documentation
Inline

## Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3609)

### DIFF
--- a/api/src/events.ts
+++ b/api/src/events.ts
@@ -23,7 +23,7 @@ export class MouseEvent {
         } else {
             // need to use screen point to convert to the map point used in creating the XY
             // however, screenX/Y output the wrong map point value which is why layerX/Y needs to be used
-            const mapPoint = mapInstance.mapI.toMap({ x: (<any>event).layerX, y: (<any>event).layerY});
+            const mapPoint = mapInstance.mapI.toMap({ x: (<any>event).layerX, y: (<any>event).layerY });
             this.xy = new XY(mapPoint.x, mapPoint.y, mapPoint.spatialReference.wkid);
         }
 
@@ -51,6 +51,7 @@ export class MapClickEvent extends MouseEvent {
     /** @ignore */
     _featureSubject: Subject<Object>;
     features: Observable<Object>;
+    apiInitiated: false;
 
     constructor(event: esriMouseEvent, mapInstance: Map) {
         super(event, mapInstance);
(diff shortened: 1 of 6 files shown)


spencerwahl:
Weren't we planning to just add a flag to the Map Click event?

Issue #3771, opened 2019-09-27
Ideas for better way to indicate search mode of grid


james-rae:
We have two different modes for column search filters on text fields.
The default mode is a "Starts with" mode that also supports `*` wild cards. 
The lazy mode is turned on via the config file and will match anywhere in the string (also supports `*`).

E.g. In default mode, `Dog` will match `Dog Park` but not `House of Dog`.  `*Dog` will match both.

There is no way on the screen to know what mode you're in.  A developer can derive it via trial and error, but a user wouldn't know about the modes.   The help content is written to suggest only the default mode exists.

Propose ways to better indicate to a user how the search works for the given mode the table is in.  This is also very helpful to the QA team, as not knowing the mode makes validating test results difficult

yileifeng:
Maybe we want to support switching between the search modes (toggling lazy filter) as a table option. This way adding it to `More Options` seems like a logical fit since it already contains different filter toggles like `Show Filters` and `Filter by extent`. It would also clarify which search mode the table is currently using.

AleksueiR:
Why do we have two search modes in the first place? Can we always just use the lazy mode? Is it more expensive to run?

It's usually not the best idea to introduce obscure modes to an application. I think it would be better to just stick with a single, most powerful search mode (maybe not in this RAMP, but in the next version).

james-rae:
Non lazy search was the original search implemented, was done as part of client requirements, and thus remains the default mode.  Changing the default in current RAMP would mean when our clients update, their site behavior will change. 

Lazy search was added much later. This is why there is two modes, and why non-lazy is default.

AleksueiR:
If we have to keep both modes, then @yileifeng 's suggestion should be fine -- add another option to the datagrid menu and use telling names to describe the difference between these two search modes, so people don't need to go to help to read up on that (they won't).

We can call them "Starts with" and "Inclusive" search modes, or something like that.

> Changing the default in current RAMP would mean when our clients update, their site behavior will change.

On the other thought, do you think anyone will notice if we swap the default search mode? 🐱‍👤

YaroRoodzit:
Tested build: v3.3.0-b3.
Works properly . According to design.

PR #3795, opened 2019-12-30
fix(enhancedTable): change default search mode

yileifeng:
## Description
Part of #3771 

Changes default filter search mode of grid to lazy. 

## Testing
Flipped the `lazyFilter` value to `false` for all previous config files with it set to `true`, while all configs that did not specify a `lazyFilter` value should default to lazy search mode.

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3795)

### DIFF
--- a/packages/ramp-core/schema.json
+++ b/packages/ramp-core/schema.json
@@ -828,7 +828,7 @@
                 },
                 "lazyFilter": {
                     "type": "boolean",
-                    "default": false,
+                    "default": true,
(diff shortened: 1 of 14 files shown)


yileifeng:
[test link](http://fgpv-app.azureedge.net/demo/users/yileifeng/default-search-mode/dist/samples/index-samples.html)

PR #3796, opened 2019-12-30
feat(enhancedTable): add table options to change filter search modes

### LINKED ISSUE #3771: Ideas for better way to indicate search mode of grid

yileifeng:
## Description
Closes #3771 

Added additional table options under table controls in header to change between filter search modes (lazy + non-lazy). 

Made this a separate PR from https://github.com/fgpv-vpgf/fgpv-vpgf/pull/3795 in case we only want one search mode, discussed [here](https://github.com/fgpv-vpgf/fgpv-vpgf/issues/3771#issuecomment-537010708)

## Testing
Under table menu/options, test the `all inclusive` and `starts with` options (used `Natural Gas Pipeline` in sample 1 for testing).

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3796)


yileifeng:
[test link](http://fgpv-app.azureedge.net/demo/users/yileifeng/search-mode-option/dist/samples/index-samples.html?sample=1)

james-rae:
___
*[packages/ramp-plugin-enhanced-table/src/index.ts, line 423 at r2](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3796#-LxRS04_AWpCOyPYfyTD:-LxRT7Bx0tKd55Ui2XNW:bj6uvdw) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/ddec28c469fb94fa67e693d7f20432db54ceea09/packages/ramp-plugin-enhanced-table/src/index.ts#L423)):*

Only developers / site configuration people would understand the lazy reference. We can remove the () text, normal user won't get it.
</blockquote></details>

Maybe we need to add more normal text to.  It's not obvious what these menu's are referring to. E.g. `All inclusive` could become `All inclusive filters`

yileifeng:
[test link](http://fgpv-app.azureedge.net/demo/users/yileifeng/search-mode-option/dist/samples/index-samples.html)

PR #3891, opened 2021-04-27
Correct french translation

### LINKED ISSUE #3886: Investigate markup in Settings Change text
In the translation file, entry key `settings.change.at`

The English entry is
```text
Change <span class="rv-control-name" translate="settings.label.{{controlName}}"></span> at:
```

But the French is only
```text
Changement {{settingName}} à:
```

Number of concerning things. Why is `span` missing in the French? Why is French using `settingName` while English is using `controlName` and deriving a separate translation key from it?

- Test to see if this even works. 
- See if the `span` is causing a different look on the English
- Figure out best fix and do that

milespetrov:
The French version contains `{{settingName}}` which doesn't print anything to the page. The English version contains a `span` translation element which is dynamically rendered based on the components value for `controlName`. 

Removing the `span` does not change the visible layout. Therefore I've replaced `{{settingName}}` with the working `span` from the English translation. 

## Before update
Screenshot from 2021-04-26 07-24-51

## After update
Screenshot from 2021-04-26 07-23-10

Closes #3886 

---
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3891)

### DIFF
--- a/packages/ramp-core/src/locales/translations.csv
+++ b/packages/ramp-core/src/locales/translations.csv
@@ -141,7 +141,7 @@ Map Navigation basemap tooltip,nav.tooltip.basemap,Basemap,1,Carte de base,1
 Metadata pane title,metadata.title,Metadata,1,Métadonnées,1
 Metadata pane expand,metadata.expand.tooltip,Expand metadata,1,Agrandir les métadonnées,1
 Hovertip loading text,maptip.hover.label.loading,Loading...,1,Chargement en cours...,1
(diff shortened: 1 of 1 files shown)


james-rae (approved):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3891)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @milespetrov)

PR #3924, opened 2021-06-03
fix(geo): optimize GeoMet requests in most cases

### LINKED ISSUE #3855: GeoMet WMS is choking the viewer when trying to consume and display layers
The GetCapabilities function calls from GeoMet are returning too much data (12+MB of XML, and growing) and this is affecting the RAMP viewer in the FGP Catalogue when viewing individual GeoMet-hosted services. This is largely due to a limitation in Esri’s JavaScript API. Changes to the Catalogue strategy for harvesting metadata from the service did not succeed. It looks like the change needs to be made to RAMP to resolve. 

From James on Slack:

> - Original code, when you make a WMS layer, the ESRI api asks the server for all its metadata and uses that to populate most of the WMS layer object.
> - When we started using the GeoMet services in CCCP (aka Tom Server), we found the metadata response was around 14 MB. Would cause "too big" errors if using a proxy, time out errors if usings CORS.
> - A new flag was added to the RAMP config to stop ESRI from getting that metadata. When the flag is on, we provide fake metadata to the WMS layer and that stops it from asking the server. Currently our fake metadata only contains the basics scraped from the layer config -- the layer id and the text name of the layer.  This was enough for the CCCP project, since it was very customized (no wizard, no default legend symbols...everything was hardcoded structured legends).
> - NrCan is now trying to consume stuff from the Tom Server and encountering similar issues. Since they are using a lot of the default stuff in RAMP, the workarounds we added for CCCP project are not good enough. In particular, because the fake metadata only has layer id and name, there is no auto-magic grabbing of the WMS legend image (while ESRI servers expose a symbol stack, WMS usually just has an image of a legend).  Note that the Tom Server can take a special parameter to return a much smaller metadata (focused on specific layers), but the ESRI api can't run that parameter, it can only ask for everything. We can make the special call with custom code.
> - I've looked at the code and the ESRI api and have a few ideas, but need some feedback from NrCan for what is feasible on their end. (edited) 
> 
> **Config Only Solution**
> - We add another optional property in config schema wmsLayerEntryNode where a legend url can be supplied.
> - NrCan adds that url value in each alternate-rcs entry
> - Our no-capabilities logic uses that url in the fake metadata
> - Possible we will need some additional changes in the code that extracts legend from the layer. Need to see if the layer structure remains the same or not.
> - Doing this will not pick up any changes to the legend url in the service. If Tom updates a legend url on the server then we’ll have a dead link until alternate-rcs gets updated
> 
> 
> **Layer Load Solution**
> - We would somehow have to make an asynch web call prior to creating the WMS layer to fetch the legend url. This pipeline is synch right now.
> - Alternative is see if we can inject the legend url after layer creation. Then we can have the asynch part be another element of the “loading” activities, and shove the result into the layer between “loading” and “loaded” phase.
> - The web call would be dealing with XML result, not JSON. Assuming our lib can handle that? We really don't make these calls in ramp. Can we use the wizards getCapabilities function (would need some modding)?
> - The code is risky, because the technique would work on Tom server but could fail on other WMS. The "layers" param on GetCapabilites is a special add on.
> - Also the "legend" structure in the XML can vary quite a bit, if memory serves. Would we code just for one expected structure? Support every possible XML spec?
> 
> Looks like any support for WMS "styles" would not work in either solution. The ESRI fake metadata class only has room for a legend url.

RyanCoulsonCA:
Resolves #3855.

This PR adds a couple of extra checks to the wmsRecord class. These checks try to determine whether the layer being requested belongs to Geomet (by looking for the `layers` parameter) and if so, creates fake metadata so the large network request is avoided (this works in a similar manner to how `suppressGetCapabilities` worked in the past). Before the layer loads, we attempt to optimize the URL using the `layers` parameter so the response from the service is smaller. When this request is fulfilled, the legend URL for the layer is updated.

For layers added via the config file: If the URL contains `geomet` and no `layers` parameter is provided but layer entries are, we attempt to add on the `layers` parameter before making any requests to the service.

A demo of this PR can be found [here](http://ramp4-app.azureedge.net/legacy/users/RyanCoulsonCA/fix-3855/samples/index-samples.html?sample=68). 

Things to check:
- [ ] The legend URL loads as expected for the WMS layer in the legend
- [ ] You can also add a WMS layer through the wizard (using an URL such as https://geo.weather.gc.ca/geomet/?lang=E&SERVICE=WMS&REQUEST=GetCapabilities&layers=CGSL.ETA_ICET)
- [ ] There are no giant XML files (~14MB) being received.

This PR was worked on by both @yileifeng and myself.

### DIFF
--- a/packages/ramp-geoapi/src/layer/layerRec/wmsRecord.js
+++ b/packages/ramp-geoapi/src/layer/layerRec/wmsRecord.js
@@ -9,7 +9,6 @@ const shared = require('./shared.js')();
  * @class WmsRecord
  */
 class WmsRecord extends layerRecord.LayerRecord {
-
     /**
      * Create a layer record with the appropriate geoApi layer type.  Layer config
      * should be fully merged with all layer options defined (i.e. this constructor
@@ -27,7 +26,7 @@ class WmsRecord extends layerRecord.LayerRecord {
         this._defaultFC = '0';
         this._featClasses['0'] = new placeholderFC.PlaceholderFC(this, this.name);
 
-        if (config.suppressGetCapabilities) {
+        if (config.suppressGetCapabilities || this.config.url.toLowerCase().indexOf('layers=') != -1) {
             this.onLoad();
         }
     }
@@ -50,7 +49,11 @@ class WmsRecord extends layerRecord.LayerRecord {
             styles: styles
         };
 
-        if (this.config.suppressGetCapabilities) {
+        // If suppressGetCapabilities is set to true, or if the URL contains the parameter `layers`,
+        // create fake metadata to prevent contacting the server.
+        if (this.config.suppressGetCapabilities || this.config.url.toLowerCase().indexOf('layers=') != -1) {
+            // TODO: if the layers provided in the URL parameter `layers` does not match what is provided in layerEntries,
+            // display an error.
             cfg.resourceInfo = {
                 extent: new this._apiRef.Map.Extent(-141, 41, -52, 83.5, {wkid: 4326}), // TODO make this a parameter post-demo
                 layerInfos: this.config.layerEntries
@@ -77,17 +80,32 @@ class WmsRecord extends layerRecord.LayerRecord {
     }
 
     /**
-    * Triggers when the layer loads.
-    *
-    * @function onLoad
-    */
+     * Triggers when the layer loads.
+     *
+     * @function onLoad
+     */
     onLoad () {
         const loadPromises = super.onLoad();
 
         const fc = new wmsFC.WmsFC(this, '0', this.config);
         this._featClasses['0'] = fc;
 
-        loadPromises.push(fc.loadSymbology());
+        // If suppressGetCapabilities is set, or the URL contains the `layers` parameter,
+        // load the data in a different manner.
+        //
+        // The `layers` parameter is used specifically for GeoMet WMS, which returns a very
+        // large amount of XML data if the parameter is not provided to specify which specific layers
+        // should be returned.
+        if (this.config.suppressGetCapabilities || this.config.url.toLowerCase().indexOf('layers=') != -1) {
+            // Attempts to retrieve the LegendURL for the layer and updates the symbology afterwards.
+            loadPromises.push(
+                this.updateCapabilities().then(() => {
+                    fc.loadSymbology();
+                })
+            );
+        } else {
+            loadPromises.push(fc.loadSymbology());
+        }
 
         Promise.all(loadPromises).then(() => {
             this._stateChange(shared.states.LOADED);
@@ -134,7 +152,7 @@ class WmsRecord extends layerRecord.LayerRecord {
                 opts.clickEvent,
                 this.config.layerEntries.map(le => le.id),
                 this.config.featureInfoMimeType)
-            .then(data => {
+            .then((data) => {
                 identifyResult.isLoading = false;
 
                 // TODO: check for French service
@@ -160,6 +178,67 @@ class WmsRecord extends layerRecord.LayerRecord {
      */
     dataSource () {
         return shared.dataSources.WMS;
+    }
+
+    /**
+     * Fetch and process metadata from WMS endpoint. Constructs a GetCapabilities call using ogc and extracts properties from resulting metadata information.
+     *
+     * @function updateCapabilities
+     * @returns {Promise} a promise that resolves once metadata properties has been processed
+     */
+    updateCapabilities () {
+        let serviceUrl = this.config.url;
+        let isGeomet = serviceUrl.toLowerCase().indexOf('geomet') != -1;
+        let hasLayersParam = serviceUrl.toLowerCase().indexOf('layers=') != -1;
+
+        // If this service is a GeoMet service and does not currently contain the `layers` parameter,
+        // build and add the parameter to the URL.
+        // NOTE: This code supports there being multiple parameters provided in layerEntries, but GeoMet does
+        // not support this yet.
+        if (isGeomet && !hasLayersParam) {
+            let layersParam = [];
+            this.config.layerEntries.forEach((entry, idx) => {
+                layersParam.push(entry.id);
+            });
+
+            if (layersParam.length > 0) {
+                serviceUrl += '&layers=' + layersParam.join(',');
+            }
+        }
+
+        // construct a ogc getCapabilities + parseCapabilities() call
+        const resPromise = new Promise((resolve, reject) => {
+            this._apiRef.layer.ogc.parseCapabilities(serviceUrl).then((data) => {
+                this.saveLegendUrls(data.layers);
+                resolve();
+            });
+        });
+        return resPromise;
+    }
+
+    /**
+     * Recursively updates the ESRI layer with the fetched legend URLs returned by getCapabilities.
+     *
+     * @function saveLegendUrls
+     * @param {Array} layers an array of layer objects returned by getCapabilities
+     */
+    saveLegendUrls (layers) {
+        layers.forEach((layer) => {
+            // Check to see if this layer belongs to the ESRI `layerInfos` array.
+            let layerInfo = this.esriLayer.layerInfos.find((item) => layer.name && item.name === layer.name);
+
+            if (layerInfo) {
+                const styles = this.esriLayer.customLayerParameters.styles;
+
+                // Set the legend URL to the style that is currently selected. If none is selected, use the default.
+                layerInfo.legendURL = typeof styles !== 'undefined' ? layer.styleToURL[styles] : layer.styleToURL[defaultStyle];
+            }
+
+            // Update sublayers if necessary.
+            if (layer.layers && layer.layers.length > 0) {
+                this.saveLegendUrls(layer.layers);
+            }
+        });
     }
 }
 
--- a/packages/ramp-geoapi/src/map/esriMap.js
+++ b/packages/ramp-geoapi/src/map/esriMap.js
@@ -112,7 +112,9 @@ function esriMap(esriBundle, geoApi) {
         }
 
         checkCorsException(url) {
-            if (corsEverywhere) {
+            // TODO: remove the check for WMS in here before merging. This is for
+            // testing purposes.
+            if (corsEverywhere || url.toUpperCase().indexOf("SERVICE=WMS") !== -1) {
                 const hostRegex = /^(?:https?:\/\/)?(?:[^@\/\n]+@)?([^:\/\n]+)/i;
                 const match = hostRegex.exec(url);
                 if (match !== null) {


yileifeng (changes requested):
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3924)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @james-rae, @RyanCoulsonCA, and @yileifeng)

___
*[packages/ramp-geoapi/src/map/esriMap.js, line 115 at r1](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3924#-MbI6LbQAzSzhMZtRWb7:-MbI6LbQAzSzhMZtRWb8:b-qbpmwr) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/17e76584588346b8a707a0e2365e4c771599907d/packages/ramp-geoapi/src/map/esriMap.js#L115)):*
> ```JavaScript
> 
>         checkCorsException(url) {
>             // TODO: remove the check for WMS in here before merging. This is for
> ```

blocking - this needs to be removed once changes are approved (needed for testing purposes)

spencerwahl (approved):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3924)* status: all files reviewed, 1 unresolved discussion (waiting on @james-rae and @RyanCoulsonCA)

james-rae:
___
*[packages/ramp-geoapi/src/map/esriMap.js, line 115 at r1](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3924#-MbI6LbQAzSzhMZtRWb7:-MbLx9GSEoL7Y8o4ZtrD:bgbsgjh) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/17e76584588346b8a707a0e2365e4c771599907d/packages/ramp-geoapi/src/map/esriMap.js#L115)):*
<details><summary><i>Previously, yileifeng (Yi Lei Feng) wrote…</i></summary><blockquote>

blocking - this needs to be removed once changes are approved (needed for testing purposes)
</blockquote></details>

as discussed, lets drop this hack, and add a geomet config snippet to sample 71 which has `corsEverywhere` turned on.

RyanCoulsonCA (commented):
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3924)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @james-rae, @spencerwahl, and @yileifeng)

___
*[packages/ramp-geoapi/src/map/esriMap.js, line 115 at r1](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3924#-MbI6LbQAzSzhMZtRWb7:-MbM1n2Y02a-xkzavW8R:b-48ivjq) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/17e76584588346b8a707a0e2365e4c771599907d/packages/ramp-geoapi/src/map/esriMap.js#L115)):*

as discussed, lets drop this hack, and add a geomet config snippet to sample 71 which has `corsEverywhere` turned on.
</blockquote></details>

Done. The new demo page can be found [here](http://ramp4-app.azureedge.net/legacy/users/RyanCoulsonCA/fix-3855/samples/index-samples.html?sample=71).

yileifeng (approved):
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3924)* status: 1 of 4 files reviewed, all discussions resolved (waiting on @james-rae, @spencerwahl, and @yileifeng)

yileifeng (approved):
> Reviewed 1 of 2 files at r1, 2 of 2 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3924)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

james-rae (changes requested):
> Reviewed 1 of 2 files at r1, 1 of 2 files at r2, 1 of 1 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3924)* status: all files reviewed, 2 unresolved discussions (waiting on @RyanCoulsonCA)

___
*[packages/ramp-core/src/content/samples/config/config-sample-71.json, line 102 at r2](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3924#-MbM2Iz_6uVMcWzRwE2V:-MbM2Iz_6uVMcWzRwE2W:b-apmfm0) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/5d5b2555836674e424c9f3529cbca974199bb09e/packages/ramp-core/src/content/samples/config/config-sample-71.json#L102)):*
> ```JSON
>         "name": "WMS Layer",
>         "layerType": "ogcWms",
>         "suppressGetCapabilities": true,
> ```

I'd rather we turn this off and add the layers param to the url. This flag already works, we are testing the `off` case.
___
*[packages/ramp-geoapi/src/layer/layerRec/wmsRecord.js, line 191 at r2](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3924#-MbMRDkA2roOGlGm4P03:-MbMRDkA2roOGlGm4P04:b1a9dkz) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/5d5b2555836674e424c9f3529cbca974199bb09e/packages/ramp-geoapi/src/layer/layerRec/wmsRecord.js#L191)):*
> ```JavaScript
>     updateCapabilities () {
>         let serviceUrl = this.config.url;
>         let isGeomet = serviceUrl.toLowerCase().indexOf('geomet') != -1;
> ```

we are fancy coders and use `!==`.  Alternately you can use `>`.

The `geomet` test is nice, but also means if those 6 letters are anywhere in the url it will declare it is geomet. maybe look for `/geomet` to narrow the hit condition? Might need to check the `/` isn't doing any weird string escaping here.
___
*[packages/ramp-geoapi/src/map/esriMap.js, line 115 at r1](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3924#-MbI6LbQAzSzhMZtRWb7:-MbM5K3DD9QFR1CxqFch:b-h0fij) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/17e76584588346b8a707a0e2365e4c771599907d/packages/ramp-geoapi/src/map/esriMap.js#L115)):*
<details><summary><i>Previously, RyanCoulsonCA (Ryan Coulson) wrote…</i></summary><blockquote>

Done. The new demo page can be found [here](http://ramp4-app.azureedge.net/legacy/users/RyanCoulsonCA/fix-3855/samples/index-samples.html?sample=71).
</blockquote></details>

drop them dirty comments too good buddy

RyanCoulsonCA (commented):
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3924)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @james-rae, @spencerwahl, and @yileifeng)

___
*[packages/ramp-core/src/content/samples/config/config-sample-71.json, line 102 at r2](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3924#-MbM2Iz_6uVMcWzRwE2V:-MbMUXQJBMr9GMUCkPD-:b-896fix) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/5d5b2555836674e424c9f3529cbca974199bb09e/packages/ramp-core/src/content/samples/config/config-sample-71.json#L102)):*

I'd rather we turn this off and add the layers param to the url. This flag already works, we are testing the `off` case.
</blockquote></details>

Done.
___
*[packages/ramp-geoapi/src/layer/layerRec/wmsRecord.js, line 191 at r2](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3924#-MbMRDkA2roOGlGm4P03:-MbMUf2M2q8XOXJe3LWB:b-oc9c5c) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/5d5b2555836674e424c9f3529cbca974199bb09e/packages/ramp-geoapi/src/layer/layerRec/wmsRecord.js#L191)):*
> we are fancy coders and use !==.  Alternately you can use >.

Very fancy. Fixed.

>The geomet test is nice, but also means if those 6 letters are anywhere in the url it will declare it is geomet. maybe look for /geomet to narrow the hit condition? Might need to check the / isn't doing any weird string escaping here.

Yeah, I was worried it could have some issues with the `geomet-beta` URL that's also used in sample 71, but it seems to work fine. I've switched it to `/geomet`.

james-rae (approved):
> Reviewed 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3924)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

spencerwahl (approved):
> Reviewed 1 of 1 files at r3, 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3924)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

james-rae (approved):
The only minor thing I've noticed is Sample 68 fails on `updateCapabilities`. You can see the layer loads on the map but the legend remains in a pre-init form, with no legend image or icon update.

If I do the GetCapabilities call (which does have the Layers parameter, so thats not the issue) through the proxy in my browser, if gives security warnings, so might be something like an expired certificate on that proxy? The network report also indicates a certificate problem.  So I'd say something to bring up at next team meeting. Not sure if we have an alternate proxy to test against.

> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3924#-:-Mbb7Qb08fBBoKI7IUzu:bj0zyy2)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

RyanCoulsonCA (commented):
Yep, this looks like a certificate problem with the proxy server. I went to the proxy link directly and clicked the `continue anyway` button and the legend icon loads fine now:



> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3924#-:-MbbZNUvEWHgNu76YLNP:b-9mncgj)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

JahidAhmed (approved):
> Reviewed 1 of 1 files at r3, 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3924)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

james-rae (approved):
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/3924)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

PR #4029, opened 2022-08-03
fix(details): only display previously selected point if it has a result

### LINKED ISSUE #4018: How to ensure that the object clicked in the map area is the same one given focus on the details pane
Hi,

Issues from our client, see description below;

***
When the user clicks on an object in the map area, they expect to see the details about that object in the details pane.  This does not work as expected.  Rather this works if no object was previously selected on the map area. If an object was selected previously, the details pane fails to forget the prior selection when the new selection is made, with a net effect of not showing the user’s selection.

Example: Select a polygon (layer A) after the map first opens and the details for that polygon will be shown.  But select another polygon (layer B) and what the user sees is “nothing found” (because there is no polygon in layer A.  The reason for this is the details pane focus remains on the layer A, and not on the layer that was just clicked (layer B).  This causes a lot of user friction.

On reflection, this is a bug and not a feature enhancement.  The user should see displayed the information on what they clicked.  What they clicked prior to doing that has no relevance, as they have moved on.
***

So if a user click on an item and nothing is present in the previous search, it should move the next layer where something is found.

RyanCoulsonCA:
Closes #4018 

This PR makes a modification to the result selection logic for the details panel. Previously, if you had selected a layer in the details results and then clicked on a point for a different layer elsewhere on the map, the previously selected layer would remain selected in the panel even if it had no results at the newly clicked point. After this PR, the previous layer will only remain selected if there exists a data point at the new spot. If no data point exists for that layer, a different result will be selected.

### DIFF
--- a/packages/ramp-core/src/app/ui/details/details.directive.js
+++ b/packages/ramp-core/src/app/ui/details/details.directive.js
@@ -87,25 +87,31 @@ function Controller($scope, $element, events, stateManager, mapService, detailSe
             }
 
             const previouslySelected = findPreviouslySelected(newValue);
-            if (previouslySelected) {
-                // pick selected item user previously selected one,
-                selectItem(previouslySelected);
-            } else if (newValue.length === 1) {
-                // or if there is a single item, pick that
+            if (newValue.length === 1) {
+                // if there is a single item, pick that
                 selectItem(newValue[0]);
             } else {
                 // otherwise, wait for the first item to get results and select that
                 deregisterFirstResultWatch = $scope.$watch(_waitForFirstResult, (result) => {
                     // check if the result returned is an object and a successfully detected first data point
                     if (typeof result === 'object' && result) {
-                        // if the user already selected an item, do not override the selection
-                        if (!self.selectedItem) {
+                        // If no layer was previously selected in the details panel, take the first point that was returned.
+                        if (!previouslySelected) {
                             selectItem(result);
+                        } else {
+                            // If there does happen to be a previous selection and that selection has a point in this result,
+                            // select that one. If it does not have a point in this result, then select the new result.
+                            if (previouslySelected.data.length > 0) {
+                                selectItem(previouslySelected);
+                            } else {
+                                selectItem(result);
+                            }
                         }
                         deregisterFirstResultWatch();
                         // otherwise the return result represents the loading status and if loading is done and all searches found nothing, no valid data point was clicked
                     } else if (typeof result === 'boolean' && !result) {
                         detailService.mApi.panels.details.header.title = 'details.label.noresult';
+                        selectItem(null);
                         deregisterFirstResultWatch();
                     }
                 });


james-rae (changes requested):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4029)* status: all files reviewed, 1 unresolved discussion (waiting on @RyanCoulsonCA and @yileifeng)

___
*[`packages/ramp-core/src/app/ui/details/details.directive.js` line 104 at r1](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4029#-N8_77k81VY-bStb7zLi:-N8_77k81VY-bStb7zLj:b-duj7xe) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/6c58f59185d8d6048ee96ef171e3f521af56d877/packages/ramp-core/src/app/ui/details/details.directive.js#L104)):*
> ```JavaScript
>                             // If there does happen to be a previous selection and that selection has a point in this result,
>                             // select that one. If it does not have a point in this result, then select the new result.
>                             if (previouslySelected.data.length > 0) {
> ```

I prob don't understand the code, but it looks like this function runs when the "first result" comes back.  So what happens on this line if the first result is from a different layer (i.e. not the previous selected layer). Would `prevSel.data.length` be 0 because results have yet to come back for the layer? 
This seems to assume we have results for the prev layer during "first result", which in theory is not guaranteed; and I don't see `findPreviouslySelected()` doing any promise stuff.
This might be the same type of problem we hit in RAMP4 with "how long to wait for prev layer if all other layers return before it"

RyanCoulsonCA:
> I prob don't understand the code, but it looks like this function runs when the "first result" comes back. So what happens on this line if the first result is from a different layer (i.e. not the previous selected layer). Would `prevSel.data.length` be 0 because results have yet to come back for the layer? This seems to assume we have results for the prev layer during "first result", which in theory is not guaranteed; and I don't see `findPreviouslySelected()` doing any promise stuff. This might be the same type of problem we hit in RAMP4 with "how long to wait for prev layer if all other layers return before it"

Yeah it seems like this is in fact a problem. I did some testing with sample 77 which has a ton of data points and clicking on the map where a few layers overlap, you can see the selected layer changes depending on which points come back first. 

I'm trying to think of a good way to deal with this... maybe we keep the previous layer selected while waiting for the results to come in (up to a certain time limit, say 500ms), and if it hasn't resolved by then we swap to the first result that comes back? 

An easier solution is obviously to not worry about previous selection at all and just simply return the first point that comes back. Not sure if this is something we want though.

Any thoughts on this would be helpful!

yileifeng (commented):
I think it makes the most sense to me to just prioritize waiting on previous selection (for promise to resolve with results or after some timeout duration) since this is intended behavior. In RAMP4, we cleared the panel on new identify requests to avoid desync but keeping the previous selection screen should be fine here.

> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4029#-:-N8iOv1Y0UgZARw2ntfC:b-achka2)* status: all files reviewed, 1 unresolved discussion (waiting on @RyanCoulsonCA and @yileifeng)

RyanCoulsonCA:
I've made some changes to the code now. It looks like the details panel doesn't actually work directly with promises, but instead works by just watching for updates to an array. 

I've now made it so if there is a previous selection, it will wait and see if any results are returned for that layer. If none are returned it checks to see if any other layers have returned a result, and if one has, then that layer is selected.

If there is no previous selection then the first result to be returned is selected as it did before.

I've done a bunch of testing with this and it seems to work as expected but let me know if you see any flaws with this logic.

james-rae (approved):
Yeah sounds like the same timing trade-offs we recently encountered in RAMP4. Either waiting too long for a slow response, or switching to another layer too early.
In the majority of cases, results are fairly snappy and this is a non-issue.
As long as we can still click a different layer while the last layer is slow-loading, that's what's critical. 

> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4029#-:-N92hf8fEBJB_RRN2qzS:bcjrghn)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA and @yileifeng)

RyanCoulsonCA (commented):
This is what I figured too. I didn't bother adding a timeout at this point because it seems like results usually come in quickly, and in the worst case results for other layers do still show up so you can always switch off.

> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4029#-:-N92ivnPD1QGmt85J7US:baw0cg7)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA and @yileifeng)

yileifeng (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4029)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

spencerwahl (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4029)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

PR #4034, opened 2022-08-26
Add support for geojson geometry collection

### LINKED ISSUE #4017: Support GeoJSON Geometry Collection
Can we add support for GeoJOSN Geometry collection. This file is valid GeoJSON but it doesn't load in the viewer.

https://raw.githubusercontent.com/jolevesq/contributed-plugins/OSDP/osdp/src/samples/geoJSON/polygons.json

yileifeng:
Closes #4017 

Added support for GeoJSON with a geometry collection at geometry level by cleaning up and converting the geometry set to a single supported geometry structure (`Point` --> `MultiPoint`, `LineString` --> `MultiLineString`, `Polygon` --> `MultiPolygon`). Multiple error checks added to ensure geometry type is consistent through the collection as well as the entire file. 

**Testing** (IPR):
- add https://raw.githubusercontent.com/jolevesq/contributed-plugins/OSDP/osdp/src/samples/geoJSON/polygons.json as GeoJSON layer and ensure no errors
- all other layer behavior should remain unchanged 

[Demo](https://fgpv-vpgf.github.io/fgpv-vpgf/4017-geometry-collection/samples/index-samples.html)

- - -
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4034)

### DIFF
--- a/packages/ramp-geoapi/src/layer.js
+++ b/packages/ramp-geoapi/src/layer.js
@@ -510,9 +510,41 @@ function validateGeoJson(geoJson) {
         fields.push({ name: oid, type: 'esriFieldTypeString' });
     }
 
+    let geomType = geoJson.features[0].geometry.type;
+    geoJson.features.forEach((feature, idx) => {
+        let featType = feature.geometry.type;
+        // handle geometry collection type case
+        if (featType === 'GeometryCollection') {
+            const geoms = feature.geometry.geometries;
+            if (geoms === undefined || geoms.length === 0) {
+                return Promise.reject(new Error('File is missing geometries for geometry collection'));
+            }
+
+            // geometry type must be consistent across entire file
+            if (geomType !== geoms[0].type && idx > 0) {
+                return Promise.reject(new Error('File contains multiple geometry types'));
+            }
+
+            // scan geometries collection to ensure that type is homogenous, throw error if not
+            geomType = geoms[0].type;
+            geoms.forEach((geom) => {
+                if (geom.type !== geomType) {
+                    return Promise.reject(new Error('File contains multiple geometry types in geometry collection'));
+                }
+            });
+
+            // clean up geojson file by converting geometry collection
+            cleanUpGeomCollection(geoJson, idx);
+
+        } else if (featType !== geomType) {
+            // geometry type must be consistent across entire file
+            return Promise.reject(new Error('File contains multiple geometry types'));
+        }
+    });
+
     const res = {
         fields: fields,
-        geometryType: geomMap[geoJson.features[0].geometry.type],
+        geometryType: geomMap[geomType],
         formattedData: geoJson,
         smartDefaults: {
             // TODO: try to find a name field if possible
@@ -523,8 +555,6 @@ function validateGeoJson(geoJson) {
     if (!res.geometryType) {
         return Promise.reject(new Error('Unexpected geometry type in GeoJSON'));
     }
-
-    // TODO optional check: iterate through every feature, ensure geometry type and properties are all identical
 
     return Promise.resolve(res);
 }
@@ -907,6 +937,41 @@ function cleanUpFields(geoJson, layerDefinition) {
     });
 }
 
+/**
+ * Convert a geometry collection to a single supported geometry structure.
+ * ([Point] --> MultiPoint, [LineString] --> MultiLineString, [Polygon] --> MultiPolygon)
+ *
+ * @function cleanUpGeomCollection
+ * @param {Object} geoJson           layer data in geoJson format
+ * @param {Number} idx               index of feature in geoJson file
+ */
+function cleanUpGeomCollection(geoJson, idx) {
+    const geoms = geoJson.features[idx].geometry.geometries;
+    // already validated type is consistent throughout non-empty collection
+    const geomType = geoms[0].type;
+    const mapping = {
+        Point: 'MultiPoint',
+        LineString: 'MultiLineString',
+        Polygon: 'MultiPolygon',
+    };
+
+    // if there is only one geometry in collection, map to single supported class
+    if (geoms.length === 1) {
+        geoJson.features[idx].geometry = {
+            type: geomType,
+            coordinates: geoms[0].coordinates,
+        };
+    } else {
+        // merge coordinates to form a multi-geometry structure
+        const combined = geoms.map(geom => geom.coordinates);
+
+        geoJson.features[idx].geometry = {
+            type: mapping[geomType],
+            coordinates: combined,
+        };
+    }
+}
+
 function makeGeoJsonLayerBuilder(esriBundle, geoApi) {
     /**
      * Converts a GeoJSON object into a FeatureLayer.  Expects GeoJSON to be formed as a FeatureCollection
@@ -1028,15 +1093,15 @@ function makeGeoJsonLayerBuilder(esriBundle, geoApi) {
         const buildLayer = () => {
             return new Promise((resolve) => {
                 // project data and convert to esri json format
-
                 const fancySR = new esriBundle.SpatialReference(targetSR);
 
+                layerId = opts.id !== undefined ? opts.id : layerId;
+                const geometryType = layerDefinition.drawingInfo.geometryType;
                 geoApi.proj.projectGeojson(geoJson, destProj, srcProj);
 
                 // terraformer has no support for non-wkid layers. can also do funny things if source is 102100.
                 // use 8888 as placehold then adjust below
                 const esriJson = Terraformer.ArcGIS.convert(geoJson, { sr: 8888 });
-                const geometryType = layerDefinition.drawingInfo.geometryType;
 
                 // set proper SR on the geometeries
                 esriJson.forEach((gr) => {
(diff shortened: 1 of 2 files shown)


james-rae:
Unfortunately it looks like any decision from the meetings wasn't recorded on the issue.
My memory says we wanted to run a filter & converter on the incoming GeoJSON. I.e. check that everything in the file was the same geometry type (so all point, all line, or all poly), then convert the content in-memory to a format that ramp could import.

This would allow it to appear in legend, have grid, identify, etc.

Notes from my markdown that I had shared with you earlier are in line with that.

- Write adapter to process any incoming GeoJSON
- Check for `GeometryCollection` as a geom `type`. Pass stuff through as is if not.
- Scan everything in the collections `geometries` array. Make sure the `type` is homogenous across the array
- If not homogenous, throw an error.
- If homogenous, convert to geom type & structure we support. Lines to `MultiLineString`, Points to `MultiPoint`, Polygons to `MultiPolygon`
- If the "collection" only had one, we could map to the non-Multi variants.
- Also need to check across the entire GeoJSON file, to ensure that everything is same type of geometry. Error if not. This check might already exist?

Might want to ping Dan and have a double-check with our NrCan friends before you make the change?

jolevesq:
@james-rae I think your solution is good. The import needs to act like a layer (legend, details and grids). Do you have some sample  JSON to test with? If not, I can ask some for you to test.

james-rae:
Great. 
It's a text file so I believe we can create some generic files that prove things work. Will just edit some standard geoJSON to have the GeometryCollection format.

If your client has more samples they would use, that would be helpful in making sure what we deliver will meet their expectations. So we won't say no to extra samples.

yileifeng (commented):
Made changes with updated [Demo](https://fgpv-vpgf.github.io/fgpv-vpgf/4017-geometry-collection/samples/index-samples.html). EDIT: testing sample files provided

> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4034#-:-NAu6Mqz0-eBOYOXO-KA:b-aahic0)* status: 0 of 4 files reviewed, all discussions resolved

james-rae (changes requested):
> Reviewed 1 of 4 files at r1, 2 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4034)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on @james-rae and @yileifeng)

___
*[`packages/ramp-geoapi/src/layer.js` line 513 at r2](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4034#-NAuDVjIAp74r96Ul6--:-NAuDVjIAp74r96Ul6-0:b6gufmx) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/b77752287313b9eecc80e6a83c34bb762952859f/packages/ramp-geoapi/src/layer.js#L513)):*
> ```JavaScript
>     }
> 
>     let geomType = geoJson.features[0].geometry.type;
> ```

should be scanning every feature, not just the first one. The sample's we're given only have one item but that does not mean there is only one.
___
*[`packages/ramp-geoapi/src/layer.js` line 947 at r2](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4034#-NAuDipmD2H51hQGno2b:-NAuDipmD2H51hQGno2c:b-tqsnyk) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/b77752287313b9eecc80e6a83c34bb762952859f/packages/ramp-geoapi/src/layer.js#L947)):*
> ```JavaScript
>  */
> function cleanUpGeomCollection(geoJson) {
>     const geoms = geoJson.features[0].geometry.geometries;
> ```

same, needs entire array processed.
___
*[`packages/ramp-geoapi/src/layer.js` line 964 at r2](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4034#-NAuECeJ7grfMUruUWh2:-NAuECeJ7grfMUruUWh3:b3lrbj1) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/b77752287313b9eecc80e6a83c34bb762952859f/packages/ramp-geoapi/src/layer.js#L964)):*
> ```JavaScript
>     } else {
>         // merge coordinates to form a multi-geometry structure
>         let combined = [];
> ```

can one-shot this pretty easy

`const combined = geoms.map(g => g.coordinates)`

yileifeng (commented):
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4034)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on @james-rae)

___
*[`packages/ramp-geoapi/src/layer.js` line 513 at r2](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4034#-NAuDVjIAp74r96Ul6--:-NC0Ji_Y3KlMs5Z_KBeG:b-3ad5ws) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/b77752287313b9eecc80e6a83c34bb762952859f/packages/ramp-geoapi/src/layer.js#L513)):*

should be scanning every feature, not just the first one. The sample's we're given only have one item but that does not mean there is only one.
</blockquote></details>

added scanning across entire file
___
*[`packages/ramp-geoapi/src/layer.js` line 947 at r2](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4034#-NAuDipmD2H51hQGno2b:-NC0JYtHCnlxr1eFpO5J:b-6afffx) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/b77752287313b9eecc80e6a83c34bb762952859f/packages/ramp-geoapi/src/layer.js#L947)):*

same, needs entire array processed.
</blockquote></details>

now only called when converting individual feature geometry collection type with `idx`
___
*[`packages/ramp-geoapi/src/layer.js` line 964 at r2](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4034#-NAuECeJ7grfMUruUWh2:-NC0JW07CZ6rkzHzW5rI:bah2568) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/b77752287313b9eecc80e6a83c34bb762952859f/packages/ramp-geoapi/src/layer.js#L964)):*

can one-shot this pretty easy

`const combined = geoms.map(g => g.coordinates)`
</blockquote></details>

Donethanks

james-rae (changes requested):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4034)* status: all files reviewed, 1 unresolved discussion (waiting on @yileifeng)

___
*[`packages/ramp-geoapi/src/layer.js` line 544 at r3](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4034#-NC0MWHW43DY5xO75foK:-NC0MWHX64s8CxxyyIAe:b-eud74a) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/6098fcd70248403830719faa2b6988554ab227ad/packages/ramp-geoapi/src/layer.js#L544)):*
> ```JavaScript
>         }
>     });
> ```

looks like you have a edge case fail with this logic. It's due to you sharing the `geomType` var both inside the GeometryCollection check, and outside at the feature level check.

Sample: Two feature file. First feature is normal feature, being a Point; Second feature is GeometryCollection, having Polygons

- `geomType = point`, set before the main loop
- loop processes first feature. Skips both `if` checks since `point === point`.
- Feature one ALL GOOD.
- loop processes second feature.
- goes into first `if` since Geom Collection
- overwrites `geomType` with first item in collection. Which is polygon.
- the inner `forEach` passes since the collection is all polygons.
- Feature two ALL GOOD.
- You've now given a gold star to a mismatch.

I'll leave it up to you how to restructure but you probably need to track a file level geometry type separately. Your inner GeomCollection loop should check all the same (as it does now), but then should also check against the file level geometry. Bit of a pain to file level the first time; sort of need to duplicate the checking logic. Or maybe you have better path.

james-rae:
*[`packages/ramp-geoapi/src/layer.js` line 544 at r3](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4034#-NC0MWHW43DY5xO75foK:-NC0Po7nBZO48uR2PpU2:b-5l8dfc) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/6098fcd70248403830719faa2b6988554ab227ad/packages/ramp-geoapi/src/layer.js#L544)):*

looks like you have a edge case fail with this logic. It's due to you sharing the `geomType` var both inside the GeometryCollection check, and outside at the feature level check.

Sample: Two feature file. First feature is normal feature, being a Point; Second feature is GeometryCollection, having Polygons

- `geomType = point`, set before the main loop
- loop processes first feature. Skips both `if` checks since `point === point`.
- Feature one ALL GOOD.
- loop processes second feature.
- goes into first `if` since Geom Collection
- overwrites `geomType` with first item in collection. Which is polygon.
- the inner `forEach` passes since the collection is all polygons.
- Feature two ALL GOOD.
- You've now given a gold star to a mismatch.

I'll leave it up to you how to restructure but you probably need to track a file level geometry type separately. Your inner GeomCollection loop should check all the same (as it does now), but then should also check against the file level geometry. Bit of a pain to file level the first time; sort of need to duplicate the checking logic. Or maybe you have better path.
</blockquote></details>

Ahhh my bad I missed a line
` if (geomType !== geoms[0].type && idx > 0) {`

Resolved

RyanCoulsonCA (approved):
> Reviewed 1 of 4 files at r1, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4034)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @yileifeng)

PR #4036, opened 2022-09-01
test: see if deployment to github works for pr's

milespetrov:
Close PR if demos build. Then check the demo is deleted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4036)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # RAMP2
 
+Let's test the new deployment works
+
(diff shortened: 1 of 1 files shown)


milespetrov:
I've pushed an update so the demo link points to samples/index-one.html. Closing to see cleanup

PR #4039, opened 2022-09-15
 layers respect defaultVisibility if not defined in config

### LINKED ISSUE #4020: Load of REST dynamic service default
ESRI REST Dynamic service may be heavy and clients publish their ESRI REST service without displaying all the default layers. It would be a useful feature if when viewer load it reads the "REST" criterion "Default Visibility" to know if it should be true or false.

RyanCoulsonCA:
Closes #4020 

Layers that do not have a visibility defined in their configuration will now set their default visibility to the value retrieved from the server.

I tested this for both dynamic and feature layers by forcing defaultVisibility to false in the source code, but if anyone knows any layers that actually have defaultVisibility set to false for testing, please post below!

You can find a demo for this PR [here](https://fgpv-vpgf.github.io/fgpv-vpgf/fix-4020/samples/index-samples.html?sample=6).

- - -
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4039)

### DIFF
--- a/packages/ramp-geoapi/src/attribute.js
+++ b/packages/ramp-geoapi/src/attribute.js
@@ -316,6 +316,7 @@ function loadServerAttribsBuilder(esriBundle, geoApi) {
                         layerData.maxScale = serviceResult.effectiveMaxScale || serviceResult.maxScale;
                         layerData.supportsFeatures = false; // saves us from having to keep comparing type to 'Feature Layer' on the client
                         layerData.extent = serviceResult.extent;
+                        layerData.defaultVisibility = serviceResult.defaultVisibility;
 
                         if (serviceResult.type === 'Feature Layer') {
                             layerData.supportsFeatures = true;
--- a/packages/ramp-geoapi/src/layer/layerRec/attribFC.js
+++ b/packages/ramp-geoapi/src/layer/layerRec/attribFC.js
@@ -63,6 +63,14 @@ class AttribFC extends basicFC.BasicFC {
         return this._layerPackage ? this._layerPackage.loadedFeatureCount : 0;
     }
 
+    get defaultVisibility() {
+        return this._layerPackage
+            ? this._layerPackage.layerData.then((data) => {
+                  return data.defaultVisibility;
+              })
+            : Promise.resolve(true);
+    }
+
     /**
      * Returns attribute data for this FC.
      *
--- a/packages/ramp-geoapi/src/layer/layerRec/dynamicRecord.js
+++ b/packages/ramp-geoapi/src/layer/layerRec/dynamicRecord.js
@@ -349,6 +349,9 @@ class DynamicRecord extends attribRecord.AttribRecord {
             }
         };
 
+        // Initialize visibility structure
+        this.leafVisibilityStruct = {};
+
         // process each leaf we walked to in the processLayerInfo loop above
         // idx is a string
         leafsToInit.forEach((idx) => {
@@ -378,6 +381,11 @@ class DynamicRecord extends attribRecord.AttribRecord {
                         dFC.extent = ld.extent;
                     }
                     dFC.extent = shared.makeSafeExtent(dFC.extent);
+
+                    // Record the defaultVisibility value for the leaf layer.
+                    dFC.defaultVisibility.then((data) => {
+                        this.leafVisibilityStruct[idx] = data;
+                    });
 
                     dFC._scaleSet.minScale = ld.minScale;
                     dFC._scaleSet.maxScale = ld.maxScale;
@@ -414,28 +422,43 @@ class DynamicRecord extends attribRecord.AttribRecord {
         });
 
         // TODO careful now, as the dynamicFC.DynamicFC constructor also appears to be setting visibility on the parent.
-        if (this._configIsComplete) {
-            // if we have a complete config, want to set layer visibility
-            // get an array of leaf ids that are visible.
-            // use _featClasses as it contains keys that exist on the server and are
-            // potentially visible in the client.
-            const initVis = Object.keys(this._featClasses)
+        const loadLayerPromise = Promise.all(loadPromises).then(() => {
+            this.config.source.layerEntries.forEach((le) => {
+                // If visibility is defined in the layer config, override the
+                // defaultVisibility value with the config value.
+                if (le.state && le.state.visibility !== undefined) {
+                    this.leafVisibilityStruct[le.index] = le.state.visibility;
+                } else {
+                    // If no state object is defined in the config, create the object and set visibility.
+                    // This is to save the value properly incase the layer is reloaded.
+                    if (!le.state) {
+                        le.state = {
+                            visibility: this.leafVisibilityStruct[le.index],
+                        };
+                    } else {
+                        // If state object is defined, but visibility is not, set visibility here.
+                        le.state.visibility = this.leafVisibilityStruct[le.index];
+                    }
+                }
+            });
+
+            // Check to see which layers are set to visible and set the visible layers.
+            const initVis = Object.keys(this.leafVisibilityStruct)
                 .filter((fcId) => {
-                    return fetchSubConfig(fcId).config.state.visibility;
+                    return this.leafVisibilityStruct[fcId];
                 })
                 .map((fcId) => {
                     return parseInt(fcId);
                 });
 
+            // If no sublayers are visible, set layer to invisible.
+            this.setVisibility(initVis.length > 0);
+
             if (initVis.length === 0) {
                 initVis.push(-1); // esri code for set all to invisible
             }
             this._layer.setVisibleLayers(initVis);
-        } else {
-            // default configuration for non-complete config.
-            this._layer.setVisibility(false);
-            this._layer.setVisibleLayers([-1]);
-        }
+        });
 
         // get mapName of the legend entry from the service to use as the name if not provided in config
         if (!this.name) {
@@ -450,7 +473,8 @@ class DynamicRecord extends attribRecord.AttribRecord {
             loadPromises.push(setTitle);
         }
 
-        Promise.all(loadPromises).then(() => {
+        Promise.all(loadPromises.concat(loadLayerPromise)).then(() => {
+            console.log('Loaded loaded');
             this._stateChange(shared.states.LOADED);
         });
     }
--- a/packages/ramp-geoapi/src/layer/layerRec/layerRecord.js
+++ b/packages/ramp-geoapi/src/layer/layerRec/layerRecord.js
@@ -298,6 +298,25 @@ class LayerRecord extends root.Root {
 
         this.extent = shared.makeSafeExtent(this.extent);
 
+        // If visibility isn't set in the config, and defaultVisibility is defined
+        // from the service, then set the visibility to be equal to the value of defaultVisibility.
+        const hasVisibleState = this.config.source.state && this.config.source.state.visibility !== undefined;
+        const defaultVisibility = this._layer.defaultVisibility;
+
+        if (!hasVisibleState && defaultVisibility !== undefined) {
+            this.visibility = defaultVisibility;
+            this.config.state.visibility = defaultVisibility;
+
+            // Save the visibility value so it is kept if the layer is refreshed.
+            if (!this.config.source.state) {
+                this.config.source.state = {
+                    visibility: defaultVisibility,
+                };
+            } else {
+                this.config.source.state.visibility = defaultVisibility;
+            }
+        }
+
         let lookupPromise = Promise.resolve();
         if (this._epsgLookup) {
             const check = this._apiRef.proj.checkProj(this.spatialReference, this._epsgLookup);


james-rae:
From the issue analysis HackMD

- Remove config defaulting for layer properties, probably `visiblity` and `opaciy`. Allow `undefined` to be a config value post-parsing
- In layer load functions, need to recognize the `undefined` value. Set some internal flag or whatnot.
- After layer metadata loads, check the flag. If set, grab server setting result and apply to layer.
- Will need to handle children of Dynamics as well as just root level layers.
- Skip for File, WFS, WMS(?)

Was any of this considered?

james-rae:
Because as is, appears to me the original problem persists: The config defaults all the booleans prior to load, so you never know if the user did or didn't provide a value

james-rae:
Or is `config.source` like an untouched/undefaulted copy?

james-rae:
*[`packages/ramp-geoapi/src/layer/layerRec/attribFC.js` line 72 at r1](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4039#-NC1FGTH75_-8e5bk96J:-NC1FGTH75_-8e5bk96K:b9sw06f) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/5f9e2d5dca9b2dc35a06487cad20f7af01ce10a2/packages/ramp-geoapi/src/layer/layerRec/attribFC.js#L72)):*
> ```JavaScript
>                   return data.defaultVisibility;
>               })
>             : true;
> ```

you prob want `Promise.resolve(true)`; right now this getter is returning Promise or boolean depending on state; might be some of your problems is because you're doing `true.then()` in reality.

james-rae:
*[`packages/ramp-geoapi/src/layer/layerRec/dynamicRecord.js` line 389 at r1](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4039#-NC1Gkd22bkCVWNZv0wy:-NC1Gkd22bkCVWNZv0wz:b-vmfq37) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/5f9e2d5dca9b2dc35a06487cad20f7af01ce10a2/packages/ramp-geoapi/src/layer/layerRec/dynamicRecord.js#L389)):*
> ```JavaScript
>                     dFC.defaultVisibility.then((data) => {
>                         if (!hasVisibleState && data !== undefined) {
>                             // This isn't working, probably not finished loading
> ```

your prob right about the loading, i'mma look into the code guts.  But at a glance your line below could just be `dFC.setVisibility(data)`

james-rae:
*[`packages/ramp-geoapi/src/layer/layerRec/dynamicRecord.js` line 428 at r1](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4039#-NC1I_Dt3G0qhiI-LwPj:-NC1I_Dt3G0qhiI-LwPk:b-gibpc1) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/5f9e2d5dca9b2dc35a06487cad20f7af01ce10a2/packages/ramp-geoapi/src/layer/layerRec/dynamicRecord.js#L428)):*
> ```JavaScript
>         });
> 
>         // TODO careful now, as the dynamicFC.DynamicFC constructor also appears to be setting visibility on the parent.
> ```

I think this block is where you need additional sauce.
We're using [this badboy](https://developers.arcgis.com/javascript/3/jsapi/arcgisdynamicmapservicelayer-amd.html#setvisiblelayers) to set the initial sublayer visibilities. Right now this code is synchronous; which was fine before since all our visible bools were known from the config. But now you have checkers above that are getting set AFTER the server spits back info about each child layer; so it happens after this code has already run.

Off the cuff idea.
This whole `if/else` gets put behind `const coulsonPromise = Promise.all(loadPromises).then(this code)`
Then you need to add `coulsonPromise` to the final `Promise.all([loadPromises, coulsonPromise])` at the bottom, to ensure the layer doesn't go to loaded until this `if/else` finishes.

Secondly, make a temporary data structure in this meatball of a function (`onLoad`) to track the boolean states of visibility. You can initialize that structure with some of the guts below (that just checks configuration as is); but then you will be updating this structure with your findings above, where you're calling `defaultVisibility()` and such.

Then it's your new structure that is used to figure out what gets passed to `_layer.setVisibility()` below.

james-rae:
*[`packages/ramp-geoapi/src/layer/layerRec/dynamicRecord.js` line 428 at r1](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4039#-NC1I_Dt3G0qhiI-LwPj:-NC1K60Q26yoqNkWlh2m:b-m8c4ao) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/5f9e2d5dca9b2dc35a06487cad20f7af01ce10a2/packages/ramp-geoapi/src/layer/layerRec/dynamicRecord.js#L428)):*

I think this block is where you need additional sauce.
We're using [this badboy](https://developers.arcgis.com/javascript/3/jsapi/arcgisdynamicmapservicelayer-amd.html#setvisiblelayers) to set the initial sublayer visibilities. Right now this code is synchronous; which was fine before since all our visible bools were known from the config. But now you have checkers above that are getting set AFTER the server spits back info about each child layer; so it happens after this code has already run.

Off the cuff idea.
This whole `if/else` gets put behind `const coulsonPromise = Promise.all(loadPromises).then(this code)`
Then you need to add `coulsonPromise` to the final `Promise.all([loadPromises, coulsonPromise])` at the bottom, to ensure the layer doesn't go to loaded until this `if/else` finishes.

Secondly, make a temporary data structure in this meatball of a function (`onLoad`) to track the boolean states of visibility. You can initialize that structure with some of the guts below (that just checks configuration as is); but then you will be updating this structure with your findings above, where you're calling `defaultVisibility()` and such.

Then it's your new structure that is used to figure out what gets passed to `_layer.setVisibility()` below.

</blockquote></details>

So in a nutshell, the flow is

1. Create an appropriate visibility tracking structure (index mapped to boolean?).
2. Init it BEFORE the `update async values` loop above. The init code is what we see below, checking `config.state.visibility` and such.
3. Then the async values code comes; almost as you have it now, but instead of setting visiblity on `dFC.setVisibility()`, you are updating your new structure.
4. The calls to `_layer.` in the `if/else` below then sit behind a promise.all that is waiting for all the other stuff (mainly the FC async calls, which are inspecting server visibility). 
5. Once that promise all resolves, we use your fancy structure to feed into the layer visibility initiation methods.  That code then resolves and our layer flips to loaded

RyanCoulsonCA (commented):
> Or is config.source like an untouched/undefaulted copy?

Yeah, as far as I can tell `config.source` is basically an un-defaulted copy, at least for feature layers. Dynamic layer leaf nodes don't seem to have this.

> From the issue analysis HackMD

Totally forgot there was a HackMD, my bad. It seems like because config.source is an unmodified version of the properties my code sort of does the same thing anyway. I will look into removing the defaults though, since it doesn't seem like dynamic leaf nodes have this property.

> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4039#-:-NC1O4h-Bez7cwHqQDzg:b-k7u56w)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @james-rae, @RyanCoulsonCA, and @yileifeng)

___
*[`packages/ramp-geoapi/src/layer/layerRec/attribFC.js` line 72 at r1](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4039#-NC1FGTH75_-8e5bk96J:-NC1Q9XJ2y_myeWNu-51:b-vlm2yz) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/5f9e2d5dca9b2dc35a06487cad20f7af01ce10a2/packages/ramp-geoapi/src/layer/layerRec/attribFC.js#L72)):*

you prob want `Promise.resolve(true)`; right now this getter is returning Promise or boolean depending on state; might be some of your problems is because you're doing `true.then()` in reality.
</blockquote></details>

Good catch :thumbsup: 
___
*[`packages/ramp-geoapi/src/layer/layerRec/dynamicRecord.js` line 428 at r1](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4039#-NC1I_Dt3G0qhiI-LwPj:-NC1Pglw9XimpKkdUJsf:b-fy0w78) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/5f9e2d5dca9b2dc35a06487cad20f7af01ce10a2/packages/ramp-geoapi/src/layer/layerRec/dynamicRecord.js#L428)):*

So in a nutshell, the flow is

1. Create an appropriate visibility tracking structure (index mapped to boolean?).
2. Init it BEFORE the `update async values` loop above. The init code is what we see below, checking `config.state.visibility` and such.
3. Then the async values code comes; almost as you have it now, but instead of setting visiblity on `dFC.setVisibility()`, you are updating your new structure.
4. The calls to `_layer.` in the `if/else` below then sit behind a promise.all that is waiting for all the other stuff (mainly the FC async calls, which are inspecting server visibility). 
5. Once that promise all resolves, we use your fancy structure to feed into the layer visibility initiation methods.  That code then resolves and our layer flips to loaded

</blockquote></details>

This is super helpful, thanks. I'm going to play around and see if I can get it working.

james-rae:
> since it doesn't seem like dynamic leaf nodes have this property

You'd probably need to use the parent layers `config.source.layerEntries[index].state.visibility` lol. I'm ok with the `.source` approach; this codebase is at a "patch only" phase of life so uber-elegance isn't required unless we need to. Your call.

james-rae:
^ a bit wrong
`config.source.layerEntries.find(le => le.index === targetIndex).state.visibility`, with maybe extra code incase the `.find` returns nothing.

RyanCoulsonCA:
I've made a bunch of changes to the PR and it works for both feature layers and dynamic layers now. Everything else seems to work the way it used to, but please look it over when you have time to make sure nothing has broken!

spencerwahl (approved):
> Reviewed 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4039)* status: all files reviewed, 3 unresolved discussions (waiting on @james-rae, @RyanCoulsonCA, and @yileifeng)

james-rae (changes requested):
> Reviewed 4 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4039)* status: all files reviewed, 5 unresolved discussions (waiting on @james-rae, @RyanCoulsonCA, and @yileifeng)

___
*[`packages/ramp-geoapi/src/layer/layerRec/dynamicRecord.js` line 472 at r2](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4039#-NCVUcvi9Bh0FqpxK-Iz:-NCVUcvi9Bh0FqpxK-J-:b23mqhj) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/2390336dd87f8eeff39a6f9025d205ba177400a0/packages/ramp-geoapi/src/layer/layerRec/dynamicRecord.js#L472)):*
> ```JavaScript
>                     this.name = serviceResult.mapName;
>                 });
>                 loadPromises.push(setTitle);
> ```

mild concern with this. The Promise.all below is inspecting the `loadPromises` array. That `.all` call happens before this runs.  I don't know if you push a promise to the array after the fact, if the `.all` will consider it.
I'd say either remove this `if` from inside the `loadLayerPromise` guts (so the push just happens in the main function, and before the final Promise.all,  or you need to wire in the `setTitle` promise to act as a resolver to `loadLayerPromise`, which is probably more work.
___
*[`packages/ramp-geoapi/src/layer/layerRec/dynamicRecord.js` line 476 at r2](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4039#-NCVVcrb488QFgfb_bUr:-NCVVcrb488QFgfb_bUs:beg24al) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/2390336dd87f8eeff39a6f9025d205ba177400a0/packages/ramp-geoapi/src/layer/layerRec/dynamicRecord.js#L476)):*
> ```JavaScript
>         });
> 
>         Promise.all([loadLayerPromise, loadPromises]).then(() => {
> ```

`loadPromises` is an array. so you're parameter is `[promise, [promise, promise, promise, ...]]`

You might want to do something like `loadPromises.concat([loadLayerPromise])`, or `[loadLayerPromise, ...loadPromises]`.

A `.push()` might work, but then your pushing `loadLayerPromise` into the array that `loadLayerPromise` is `.all`ing on which might create a black hole or something.
___
*[`packages/ramp-geoapi/src/layer/layerRec/layerRecord.js` line 300 at r2](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4039#-NCVXbY6-f8lCf30w44C:-NCVXbY6-f8lCf30w44D:b-x8nkad) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/2390336dd87f8eeff39a6f9025d205ba177400a0/packages/ramp-geoapi/src/layer/layerRec/layerRecord.js#L300)):*
> ```JavaScript
> 
>         this.extent = shared.makeSafeExtent(this.extent);
>         this.defaultVisibility = this._layer.defaultVisibility;
> ```

this is becoming a layer property. I don't see it being read anywhere other than this function. It also has same name as the FC property but different data type (boolean vs Promise(boolean)).

If we don't need it outside of here, can this just be a `const`?
If we do need it persisted, can we add a comment about its type and whatnot. Without typescript this will get confusing.

RyanCoulsonCA (commented):
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4039)* status: 4 of 6 files reviewed, 5 unresolved discussions (waiting on @james-rae, @RyanCoulsonCA, @spencerwahl, and @yileifeng)

___
*[`packages/ramp-geoapi/src/layer/layerRec/dynamicRecord.js` line 472 at r2](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4039#-NCVUcvi9Bh0FqpxK-Iz:-NCVbZU68seRgrK1q9ie:b-zcyad4) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/2390336dd87f8eeff39a6f9025d205ba177400a0/packages/ramp-geoapi/src/layer/layerRec/dynamicRecord.js#L472)):*

mild concern with this. The Promise.all below is inspecting the `loadPromises` array. That `.all` call happens before this runs.  I don't know if you push a promise to the array after the fact, if the `.all` will consider it.
I'd say either remove this `if` from inside the `loadLayerPromise` guts (so the push just happens in the main function, and before the final Promise.all,  or you need to wire in the `setTitle` promise to act as a resolver to `loadLayerPromise`, which is probably more work.
</blockquote></details>

You're right, didn't catch that. I've moved it outside of the loadLayerPromise.
___
*[`packages/ramp-geoapi/src/layer/layerRec/dynamicRecord.js` line 476 at r2](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4039#-NCVVcrb488QFgfb_bUr:-NCVbea30Zky_SRFXcOK:bt2vr0y) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/2390336dd87f8eeff39a6f9025d205ba177400a0/packages/ramp-geoapi/src/layer/layerRec/dynamicRecord.js#L476)):*

`loadPromises` is an array. so you're parameter is `[promise, [promise, promise, promise, ...]]`

You might want to do something like `loadPromises.concat([loadLayerPromise])`, or `[loadLayerPromise, ...loadPromises]`.

A `.push()` might work, but then your pushing `loadLayerPromise` into the array that `loadLayerPromise` is `.all`ing on which might create a black hole or something.
</blockquote></details>

Fixed!
___
*[`packages/ramp-geoapi/src/layer/layerRec/layerRecord.js` line 300 at r2](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4039#-NCVXbY6-f8lCf30w44C:-NCVbfGI-i_S-XdbylpW:b-28bwbm) ([raw file](https://github.com/fgpv-vpgf/fgpv-vpgf/blob/2390336dd87f8eeff39a6f9025d205ba177400a0/packages/ramp-geoapi/src/layer/layerRec/layerRecord.js#L300)):*

this is becoming a layer property. I don't see it being read anywhere other than this function. It also has same name as the FC property but different data type (boolean vs Promise(boolean)).

If we don't need it outside of here, can this just be a `const`?
If we do need it persisted, can we add a comment about its type and whatnot. Without typescript this will get confusing.
</blockquote></details>

It definitely doesn't need to be a layer property. I've changed it to be a const.

yileifeng (approved):
> Reviewed 2 of 6 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4039)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

spencerwahl (approved):
> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4039)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

Issue #4066, opened 2022-10-27
Feature Layer Initial Filter is not working


james-rae:
The config defaulter is not passing along the property `initialFilteredQuery` for feature layer configs, so it never makes it to the actual ESRI layer.

Also this property is missing from the config schema.

PR #4067, opened 2022-10-27
Feature layer filter fix

james-rae:
Donethankses #4066

Makes the config defaulter pass along the filter value so the layer can use it. Also added the property to the schema.

[Demo page](https://fgpv-vpgf.github.io/fgpv-vpgf/defexp/samples/index-fgp-en.html)

Then paste into console

```text
var map = RAMP.mapInstances[0];

var lconf = {
	"id": "funtimes",
	"layerType": "esriFeature",
	"state":{"visibility": true},
	"url": "https://maps-cartes.ec.gc.ca/arcgis/rest/services/EcoGeo/EcoGeo/MapServer/6",
	"initialFilteredQuery": "Year_='2009'"
}

map.layers.addLayer(lconf)
```

When the layer draws, there will be less points than usual.
As is tradition, it appears the original specs of this item are lost, but seems to me this filter only gets applied initially; any subsequent filter stuff erases it. So you can toggle visibility and get the full set of points to appears.
Also the grid does not seem to be respecting the filter. We could log this, but seeing as this option has been broken for years, I think this is more a "make it work in RAMP4" initiative.

- - -
This change is [Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4067)

### DIFF
--- a/packages/ramp-core/schema.json
+++ b/packages/ramp-core/schema.json
@@ -1153,6 +1153,11 @@
                         "$ref": "#/definitions/fieldMetadataEntry"
                     },
                     "description": "Specifies options for the fields of a layer."
+                },
+                "initialFilteredQuery": {
+                    "type": "string",
+                    "default": "",
+                    "description": "A where clause filter to be initially applied to the layer"
                 }
             },
             "required": ["id", "layerType", "url"],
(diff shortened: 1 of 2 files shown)


yileifeng (approved):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4067)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

RyanCoulsonCA (approved):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4067)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @james-rae)

Issue #4087, opened 2023-01-12
Remove HTTP URL trickery


james-rae:
Parts of the RAMP core will generate certain URLs using the protocol that the host site is in. In particular, the URL to the ESRI JS API and the ESPG projection lookup service (if no urls are supplied in the config, which is typical case).

Since `http` is now becoming less common, and CORS is more common, we are finding this url logic can cause CORS errors when attempting to access the `http` version of those services. Most notably when attempting to run RAMP in develoment mode using `localhost`. We could probably config our `rush` scripts to serve via `https` but this could also happen in any `http` host sites, as well it would not be shocking if those two services depreciated their `http` endpoints.

Remove the trickery. Always use the `https` version of those URLs.

- `epsg.feature.ts`
- `config.service.js`
- `global.registry.js` (not used but clean it up anyways).

Note there is also some stuff in `loader-service.directive.js`, but I think we should leave this as is, since the affected URL is user supplied and `http` may be valid in niche scenarios.

prosperva:
I support this one @james-rae.  I have had to create a localhost certificate to solve CORS issues.

PR #4088, opened 2023-01-16
Remove Protocol Matching, Other Small Fixes

james-rae:
Donethankses #4087 

Removes internal logic that attempts to protocol match (`http`, `https`) when constructing urls to ESRI JS API and EPSG lookup. Will now always request the `https` version.

Updates config schema to indicate the `geoSuggest` url (part of the geosearch) is not used. Also removes an internal check that wanted it to exists. Sample 1 has the url removed from the config to prove the app will start and geosearch as usual.

Updates all the samples to reflect the latest domain change in the lambert basemaps. The redirect between the old domain and new domain were causing CORS problems.

### DIFF
--- a/packages/ramp-core/api/src/ui.ts
+++ b/packages/ramp-core/api/src/ui.ts
@@ -189,7 +189,7 @@ export class Basemap {
  *         {
  *             "id": "CBMT",
  *              "layerType": "esriFeature",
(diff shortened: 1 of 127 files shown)


RyanCoulsonCA (approved):
> Reviewed 127 of 127 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4088)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @yileifeng)

PR #4094, opened 2023-02-27
feat: add custom stopPropagation & preventDefault

milespetrov:
Adds a way to bypass the focus manager when attempting to call `stopPropagation`  and `preventDefault` on events originating from inside the map. Since this is adding more options to an event nothing should be breaking.

This PR is for E2Reg where ramp is placed inside a form and hitting the `Enter` key on active inputs submits the form/page. By calling `originalPreventDefault` and `originalStopPropagation` we can stop that event from moving past ramp.

### DIFF
--- a/packages/ramp-core/src/app/focus-manager.js
+++ b/packages/ramp-core/src/app/focus-manager.js
@@ -905,6 +905,9 @@ HTMLElement.prototype.focus = $.fn.focus = function () {
     }
 };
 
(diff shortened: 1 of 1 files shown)


james-rae (approved):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fgpv-vpgf/fgpv-vpgf/4094)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @milespetrov)

